Issue #11, opened 2021-03-08
static highlighting


pietroppeter:
there's no really need to add javascript to highlight stuff at runtime. I should be able to do it when rendering html.
there was some discussion today in IRC #science channel (which unfortunately is not logged).

Yardanico:
Some options:

- https://github.com/alecthomas/chroma - a tool in Go, used by Hugo for syntax highlighting
- https://nim-lang.org/docs/highlite.html - pure Nim, but only supports a few languages, example of usage - https://github.com/nim-lang/Nim/blob/devel/lib/packages/docutils/rstgen.nim#L992 (from rstgen which is used for nim doc generation and on the nim forum)
- Use highlight.js itself - make some lightweight JS CLI app (maybe custom-built), will require Node 

There are many other options of course

Vindaar:
Just adding `Pygments`, because aside from the Python dependency it makes for a good alternative:

https://pygments.org/

[Mini example](https://pygments.org/demo/?lexer=nim&code=import%20math%2C%20unchained%0Alet%20R_Earth%20%3D%206371.km%0A%0Aproc%20distanceAtmosphere(%CF%91%3A%20Radian%2C%20d%3A%20KiloMeter%20%3D%2036.6149.km)%3A%20UnitLess%20%3D%0A%20%20result%20%3D%20sqrt((R_Earth%20%2F%20d%20*%20cos(%CF%91))%5E2%20%2B%202%20*%20R_Earth%20%2F%20d%20%2B%201)%20-%20R_Earth%20%2F%20d%20*%20cos(%CF%91)%0A)

Wouldn't use it as a default to not have it as a dependency, but it might be nice as an alternative. I use it when exporting Org files to LaTeX and HTML. I like it, because of the large number styles it ships with and in my experience it just works.

I've never used it as a library though, so I don't know how easy it is to interface with from that point of view.

Yardanico:
@Vindaar the chroma tool I linked actually uses pygments definitions under the hoog

Vindaar:
> @Vindaar the chroma tool I linked actually uses pygments definitions under the hood - "Chroma is based heavily on Pygments, and includes translators for Pygments lexers and styles."

Oh, haha. I thought it was a pure Go solution.

Yardanico:
> > @Vindaar the chroma tool I linked actually uses pygments definitions under the hood - "Chroma is based heavily on Pygments, and includes translators for Pygments lexers and styles."
> 
[...]

PR #14, opened 2021-03-08
Use Nim's highlite module for Nim syntax highlighting

Yardanico:
As discussed in https://github.com/pietroppeter/nimib/issues/11. The only somewhat notable missing highlighting compared to highlight.js is that pragmas are not highlighted. And of course, the code wasn't thoroughly tested :)

### DIFF
--- a/src/nimib/highlight.nim
+++ b/src/nimib/highlight.nim
@@ -7,7 +7,7 @@ func nimNormalize(s: string): string =
   # Copied from strutils.normalize
   # Normalizes a Nim identifier to all lowercase:
   # - All letters except the first one are lower-cased
-  # - Underscores are ignores
+  # - Underscores are ignored
   result = newString(s.len)
   var j = 0
   for i in 0 .. len(s) - 1:
--- a/src/nimib/highlight.nim
+++ b/src/nimib/highlight.nim
@@ -1,0 +1,94 @@
+import std/strutils
+from std/cgi import xmlEncode
+
+import packages/docutils/highlite
+
+func nimNormalize(s: string): string =
+  # Copied from strutils.normalize
+  # Normalizes a Nim identifier to all lowercase:
+  # - All letters except the first one are lower-cased
+  # - Underscores are ignores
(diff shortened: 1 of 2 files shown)


Yardanico:
This PR doesn't have the necessary changes to disable highlight.js though

src/nimib/highlight.nim
```diff
@@ -0,0 +1,94 @@
+import std/strutils
+from std/cgi import xmlEncode
+
+import packages/docutils/highlite
+
+func nimNormalize(s: string): string =
```
pietroppeter, line 6: what's the difference between this and https://nim-lang.org/docs/strutils.html#normalize%2Cstring ?

src/nimib/highlight.nim
```diff
@@ -0,0 +1,94 @@
+import std/strutils
+from std/cgi import xmlEncode
```
pietroppeter, line 2: wow, I did not know we have a cgi module! :)

pietroppeter:
thanks, I will check it later and see if there is something to be done before merging, but at the moment it looks good! Once I also rerun ptest with this we will also have some testing coverage.

src/nimib/highlight.nim
Yardanico: @pietroppeter as explained in the comments, my modification actually compares nim identifiers, while that proc says "This should NOT be used to normalize Nim identifier names."

The difference is when we are comparing the first character of the string, it's case sensitive in Nim

pietroppeter: ok, I see it now, the difference with the strutils version is the `i == 0` branch that keeps case of first character (something that we need in fact).

pietroppeter (approved):
all good I will now pull this locally, apply the patches to removed default highlight.js, update README and merge

pietroppeter:
doing the above process I ended up creating a new PR (#15) with additional commits. In the end merging #15 happened to merge also #14 at the same time, not sure why but all good. Thanks again @Yardanico!

PR #15, opened 2021-03-08
Add static highlight

pietroppeter:
same as PR #14 with some added changes. I was supposed to update that PR but I made a mistake in branch name and so I had to create a new PR. Still commits by @Yardanico are still correctly author so I will merge this and close the other.


pietroppeter:
I forgot to update README (description on how highlight is done), I will do it later when I revise the whole README

PR #21, opened 2021-03-24
Use path to find Git

Yardanico:
Right now `nimib` will fail because it won't be able to find `git` (unless it's somehow already available in the environment). poUsePath will search for the binary in the PATH (which should be the default)

Reported by @ajusa on Discord and then confirmed by me.

### DIFF
--- a/src/nimib/gits.nim
+++ b/src/nimib/gits.nim
@@ -3,15 +3,15 @@ import std / [os, osproc, strutils]
 import nimib / paths
 
 proc isGitAvailable*: bool =
-  execCmdEx("git --version", options={}).exitcode == 0
+  execCmdEx("git --version", options={poUsePath}).exitcode == 0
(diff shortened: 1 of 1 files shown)


pietroppeter:
Thanks for this! I knew that `gits` module was kind of hacky but for some reason I cannot reproduce the bug in my system (maybe a difference when using Windows?). I do need to add a CI that might catch these kinds of things.

Yardanico:
@pietroppeter I think it's due to the fact that `poUsePath` has no effect on Windows (it always searches the PATH)

Yardanico:
@pietroppeter it's actually documented in https://nim-lang.org/docs/osproc.html#ProcessOption - 
```
## Asks system to search for executable using PATH environment
              ## variable.
              ## On Windows, this is the default.
```

PR #22, opened 2021-03-24
Support SSH GitHub remotes - convert them into normal github URLs

Yardanico:
This is based on https://github.com/pietroppeter/nimib/pull/21 so merge it first :)

A "default" GitHub SSH remote looks like this: `git@github.com:Yardanico/nimib.git`

Right now I only decided to support the "default" GitHub SSH remotes (the ones that it gives when creating a new repo or checking out an existing one) without the `ssh://` prefix.


pietroppeter:
yep, as I mentioned in #21, it was definitely hacky and apparently full of gotchas :D
Thanks again!

PR #31, opened 2021-03-28
Review readme

pietroppeter:
reviews readme and adds CI. Also renamed nbProjDir to nbHomeDir


pietroppeter:
why does the CI runs twice?

Yardanico:
@pietroppeter because `push` means any push for any branch. See https://docs.github.com/en/actions/reference/events-that-trigger-workflows

Yardanico:
I guess you'd want to configure `push` CI to only run in the `main` branch as shown in the example in the link

pietroppeter:
:D thanks. I took the CI set up somewhere. Makes sense to remove one of the two or there are cases where they are both needed? I cannot tell from documentation

Yardanico:
@pietroppeter it's fine to have both, but it's better for push to specify that it only should run in the main branch, so something like
```yaml
name: Run tests
on: 
  push:
    branches:
      - main
  pull_request:
jobs:
  tests:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: jiro4989/setup-nim-action@v1.1.4
        with:
          nim-version: '1.4.4'
      - run: nimble deps
      - run: nimble tdeps
      - run: nimble ptest
```

Untested but kinda should work :P

pietroppeter:
well, I leave them both. I was about to commit your untested version but I am kinda scared by `pull_request:` without nothing after (is it valid yaml? will it give problems?)... already struggling enough with this CI :)

Yardanico:
@pietroppeter yes it's valid YAML and specified explicitly in the link I gave

Yardanico:
> If you need to specify activity types or configuration for an event, you must configure each event separately. **You must append a colon (:) to all events, including events without configuration.**

pietroppeter:
yeah, finally green! I am so happy I will also try that!

PR #35, opened 2021-05-03
Refactor renders (fix #24)

### LINKED ISSUE #24: refactor rendering
a big refactor in rendering is ongoing (branch [refactor-render](https://github.com/pietroppeter/nimib/tree/refactor-renders)).

I will add render objects and separte the rendering part from the block part. blocks will be generic (no block kind anymore) to simplify creation of custom blocks. also it will simplify how to add or customize a rendering backend.

Basically every block will have only a renderPlan (just a seq of strings). and according to backend it will execute the different steps. block renderPlan can be customized and rendering backend also will be customizable.

pietroppeter:
wip, fix #24

### DIFF
--- a/src/nimib/blocks.nim
+++ b/src/nimib/blocks.nim
@@ -23,19 +23,14 @@ template manageErrors*(identBlock, body: untyped) =
 template nbTextBlock*(identBlock, identContainer, body: untyped) =
   # assume body is a string
   identBlock = newBlock(nbkText, toStr(body))  # nbDoc.flags.update(flags)
-  identBlock.renderPlan = @["mdToHtml"]
+  identBlock.renderPlan = nbTextBlockDefaultSteps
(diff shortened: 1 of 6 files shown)


src/nimib/blocks.nim
```diff
@@ -7,9 +7,11 @@ macro toStr*(body: untyped): string =
 proc newBlock*(kind: NbBlockKind, code: string): NbBlock =
   # I cannot use this directly in nbBlocks (nbText, nbCode, ...)
   # or it will substitute kind and body fields with their values
-  NbBlock(kind: kind, code: code)
+  result = NbBlock(kind: kind, code: code)
+  result.context = newContext(searchDirs = @[])  # searchDirs=@[] gives error, why?
```
pietroppeter, line 11: because @= is understood as an indentifier

PR #44, opened 2021-05-14
create dir if it doesn't exist

HugoGranstrom:
It became quite simple, we basically create the dir if it doesn't exist

### DIFF
--- a/src/nimib/docs.nim
+++ b/src/nimib/docs.nim
@@ -1,7 +1,11 @@
+import std/os
 import browsers
 import types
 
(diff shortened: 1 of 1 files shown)


pietroppeter:
Thanks, looks good! I expect createDir to also works in case of nesting non existing directories (e.g. `none/of/these/exist/`), right?

HugoGranstrom:
> I expect createDir to also works in case of nesting non existing directories

Yes it does :) From Nim Docs:

> The directory may contain several subdirectories that do not exist yet. The full path is created

Issue #25, opened 2021-03-28
use a nb global object


pietroppeter:
I think I will want to have a single nb global object that contains all variables that are now put in the global space by nbInit. Still not too sure about this.

pietroppeter:
Possibile advantage: I could have a `nb.md` global object that will be the markdown converter (customizable and initialized only once)

pietroppeter:
This means that it instead of having multiple global objects injected in scope by `nbInit` (eg `nbDoc`, `nbBlock`, `nbThisFile`, `nbHomeDir`, ...) we have a single `nb` object with appropriately defined fields (eg `nb.doc`, `nb.block`, ...). This would break current internal api, but it would better track what is done inside nbInit and help with customizations.

pietroppeter:
reflecting on this I think that instead of a new type for a `nb` global object, I can probably update `NbDoc` type to incorporate all objects that are injected globally and only change the name of inject object from `nbDoc` to `nb` (and add a `nbDoc` template as an alias for `nb` to reduce api breaking).

HugoGranstrom:
That makes a lot of sense. If we can make it into a single injected variable it would make it easier for users to remember where to change stuff 😄

PR #45, opened 2021-05-14
Alternative init

Clonkk:
@pietroppeter  This was what I was talking about.

For reference it would allow to do things like : 

myfile_custominit.nim
```nim
import nimib

proc nbUserInit*(nbThisDir, nbHomeDir: AbsoluteDir, nbThisFile: AbsoluteFile, nbUser: string) : NbDoc =
  echo "This is my custom init BABY"
  # Do default init
  var nbDoc = initNbDoc(nbThisDir, nbHomeDir, nbThisFile, nbUser) # Optionnal since you can also initialize NbDoc from scratch
  let nbSrcDir = nbHomeDir / "../books".RelativeDir
  nbDoc.filename = (changeFileExt(nbThisFile, ".html").relativeTo nbSrcDir).string
  nbDoc.context["here_path"] = (nbThisFile.relativeTo nbSrcDir).string
  nbDoc.context["title"] = nbDoc.context["here_path"]
  nbDoc.partials["header_center"] = "<code>" & nbDoc.context["title"].castStr & "</code>"
  nbDoc.context["home_path"] = (nbSrcDir.relativeTo nbThisDir).string
  return nbDoc
```

and in basic_plotting.nim : 
```nim
import nimib
import myfile_custominit
import ggplotnim

nbInit(NbCustomInit) # Call Custom init
# nbInit() # This version would call normal init

# ... Rest of thefile
```

That way you don't rely on a compile define hidden inconfig file hard-coded file name or injected variable that aren't declared anywhere from the user perspective.

I mark this as a draft because there is more work to be done to polish the init remove the would-be obsolete code. If you want I can target a branch other than ``main`` and we can take it from there

### DIFF
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -8,8 +8,25 @@ export defaults.useLatex, defaults.darkMode
 from mustachepkg/values import searchTable, searchDirs, castStr
 export searchTable, searchDirs, castStr
 
+type
+  NbCustomInit* = object
+  NbDefaultInit* = object
 
-template nbInit*() =
+proc initNbDoc*(nbThisDir, nbHomeDir: AbsoluteDir, nbThisFile: AbsoluteFile, nbUser: string) : NbDoc =
+  var nbDoc : NbDoc
+  nbDoc.author = nbUser  # never really used it yet, but probably could be a strdefine
+  nbDoc.filename = changeFileExt(nbThisFile.string, ".html")
+  nbDoc.render = renderHtml
+  nbDoc.templateDirs = @["./", "./templates/"]
+  nbDoc.partials = initTable[string, string]()
+  nbDoc.context = newContext(searchDirs = @[])
+  nbDoc.context["home_path"] = (nbHomeDir.relativeTo nbThisDir).string
+  nbDoc.context["here_path"] = (nbThisFile.relativeTo nbHomeDir).string
+  nbDoc.context["source"] = read(nbThisFile)
+  defaults.init(nbDoc)
+  return nbDoc
+
+template nbInit*(t: untyped = type NbDefaultInit) =
   # I think I want to migrate to a single global object nb
   # with nb.doc as nbDoc and nb.block (or nb.blk?) as nbBlock
   # the global object will also contain all those paths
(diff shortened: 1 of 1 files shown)


pietroppeter:
just to recover context (too many threads, I am getting lost*! :)), this is the discussion: https://github.com/SciNim/getting-started/issues/9#issuecomment-841146566

As mentioned there, the reason I wanted an "invisible" customization mechanism is that it helps to pass a bunch of documents from a default configuration to a custom configuration. And the api of the single document (which is actually the content of the "page") remains very clean and helps separate a bit more the presentation layer from the content layer. I still think it is a valid mechanism and it could be even made more stealth (without depending on define switches, see answer in issue above, will need to create a specific issue in nimib).

But an alternative and more visible mechanism for customization could be a good idea too (I also though about something like this, before going in this sort-of-odd custom include direction), especially if the two mechanism can coexist for a while. Being nimib very young, I think we should experiment a bit with the api and see what survives and what works best. At some point we might need to deprecate stuff.

I just gave a quick look at the code and I like how it looks.
Just one remark, I have the idea at some point to create a global `nb` object to refactor all this variables injected in the global space in a single object (to clean up and track better what changes), see https://github.com/pietroppeter/nimib/issues/25
I suspect that refactoring could actually help to implement a feature like this. What do you think of #25?

*actually I am very happy about all these discussion and hopefully this weekend I will revise all the loose points.

pietroppeter:
after losing a bit of time understanding whether or not the custom include was included (it was, the bug was elsewhere), this feature is growing a lot on me! We should have it!

Clonkk:
Regarding Global object, it basically enforce that you only have one page per Nim file.

Which IMHO, is not a bad thing as long as it's documented clearly - I would find having multiple document per Nim file messy anyway.

pietroppeter:
> Regarding Global object, it basically enforce that you only have one page per Nim file.

I am not sure how that differs from what we have today. We do have `nbDoc`, `nbBlock` global objects that make it difficult to have more than one page per Nim file. I do need to precise better what `nb` global object means though. I will do that in the issue.

Clonkk:
I'll close this as it was just an example.

We know where to find it if we want to refactor the init this way.

PR #43, opened 2021-05-14
Mdbook fix #40

pietroppeter:
WIP

- [X] convert [`index.hbs` from mdbook](https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs) to a `document.mustache`
- [X] add to `book` folder most of the content of mdbook [src/theme](https://github.com/rust-lang/mdBook/tree/master/src/theme) folder (but moved all css to folder css and moved all js to a js folder)
- [X] add a placeholder `book.nimble` to set the home folder.
- [X] add a `nbPostInit.nim` that will provide the configuration of the book. made it runnable so that it can generate a test page
- [x] fix and understand `theme_option` and `theme_default` so that `nbPostInit` renders nicely
- [x] fix FontAwesome not loading correctly in `nbPostInit`
- [x] add manually content to `toc.mustache`
- [x] add some sample content (chapters, sections, ...)

PR #47, opened 2021-05-14
add minimal mdbook

pietroppeter:
replaces #43

* minimal mdbook copied and adapted from rust mdbook.


pietroppeter:
nope. conflicts because I forgot to update main before branching... replaced by #48

Issue #34, opened 2021-05-03
nbFile and nbShell


pietroppeter:
`nbFile`: write nim code content into a file that might be compiled/run later
`nbShell`: run shell command and capture output and error

Example:

```
nbFile("module.nim"): # write the following content to module.nim file
  proc hello*(name: string) = echo "hello " & name

nbFile("main.nim"): # write the following content to main.nim file
  import module
  hello "you"

nbShell("nim r main") # run main.nim and captures output
```

this could also be used for creating nim files that can be compiled with a different backend (e.g. javascript). example from karax:

```
nbFile("hello.nim"):
  include karax / prelude
  
  proc createDom(): VNode =
    result = buildHtml(tdiv):
      text "Hello World!"
  
  setRenderer createDom

nbShell("karun -r helloworld.nim")
```

* in html output the code block should show the name of the file (possibly also a different styling?).
* nbShell should be styled differently from standard nim code block
* not sure what the advantage of using nim code instead of a string, since the nim code will not checked by tools like `nim check` or `nimsuggest` (basically code will be discarded and only its string representation will be kept).

Clonkk:
The nbFile functionality is a great idea and would help me write the tutorial on interacting with Python and Julia. 

Are you already working on it ?

pietroppeter:
No, not working on it. The idea for me was to finish the PR on refactoring NbBlock and rendering, but I have had really little time last few months (hopefully this situation is changing). I do not see any problem if you want to already give it an implementation (thanks!) and I will refactor that when the time comes.

Clonkk:
Do you intend to keep the file in the book ?

Clonkk:
I'll keep the file for now, so it's stays simple. It's easy enough to remove manually if need be

pietroppeter:
realized now I never answered to this, sorry. Anyway, I agree behavior for nbFile should be to keep the file, we could at some point provide some more ergonomic system for cleaning up files.

HugoGranstrom:
A use-case I've stumbled upon when trying some things with NimiSlide:

[...]

PR #59, opened 2021-06-20
add `nbCodeBlock` to emit code in a `block` statement

Vindaar:
I started writing some first tutorials for `getting-started`. @HugoGranstrom gave me some feedback that for reading purposes it's better to only reference variables defined in the current code block.

With that in mind I realized that it'd be nice to just emit code into a `block` statement so that one doesn't have to write variables `x1` to `xN`. 

I added a `nbCodeBlock` (which I realized is the name internally used as well :see_no_evil:) and a static flag that emits it into a `block` in that case. Seems to work as I expected. So:

```nim
nbCodeBlock:
  let x = 5
nbCodeBlock:
  let x = 10
```
is valid, whereas of course
```nim
nbCode: 
  let x = 5
nbCode:
  let x = 10
```
still throws a redefinition error.

### DIFF
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -66,6 +66,11 @@ template nbInit*() =
   template nbCode(body: untyped) =
     nbCodeBlock(nbBlock, nbDoc, body)
 
+  template nbCodeInBlock(body: untyped) =
+    block:
+      nbCode:
+        body
+
   template nbImage(url: string, caption = "") =
     if isAbsolute(url) or url[0..3] == "http":
       # Absolute URL or External URL


pietroppeter:
I would actually go with:
```nim
block:
  nbCode:
    let x = 1
block:
  nbCode:
    let x = 2
```

Vindaar:
Yes, Hugo mentioned that afterwards as well. Seems like a common use case that could use an overload, don't you agree?

HugoGranstrom:
An alternative Vindaar proposed if you think this PR doesn't add enough new: https://discord.com/channels/371759389889003530/755344160592101389/856121013595275275

TL;DR: the `block: nbCode: ` approach doesn't work with imports so isolated/independent code blocks can't be written at the moment without first doing the imports.

pietroppeter:
> Seems like a common use case that could use an overload, don't you agree?

I think I would keep using the `nbCode` wrapped in `block` explicitly, it would be anyway clearer for me to use (and if I really use it that often I would create a template for that).

For the use case of having imports in a scoped block, I think the answer will come when adding API to create and run a temporary file with the given code content  (which could span one or more blocks), along the lines of #34.

Regarding adding this, we could, but I would change two things:
* rename `nbCodeBlock` to `nbCodeInBlock`
* change implementation to the template one without changing signature of standard `nbCode`:

```nim
template nbCodeInBlock(body: untyped) =
  block:
    nbCode:
      body
```
this implementation does something a bit different, but for the use case it is actually the same.

Vindaar:
Yes, this is what I ended up using here:

https://github.com/SciNim/getting-started/pull/22

I think it would be useful to have so we don't have to define it in every file we want to use it. I'll change the implementation to this.

Issue #51, opened 2021-05-27
Custom output directory as command line flag


HugoGranstrom:
This is something that we have had to do workarounds for in nimiBook, the fact that we don't want the output `.html` and images to be beside the `.nim` files. In that case, we would like to be able to set an output directory relative to the Nimble project but it still feels a bit like jumping through hoops because we must make sure the srcDir, homeDir, etc match up to what we expect. (may not be anymore though:P) 

And this is also useful for nimiBoost so that one can compile and preview single `.nim` files that are located in a nimble project without having to care about the fact that it is in a nimble directory. I, therefore, suggest that we find a way to make it easier to choose these in the compile command already. 

The needs I have for the nimiBoost extension is to set the output dir primarily, if it has an ugly name in the header I don't care about but if it too could be fixed that would be great.

From my research the easiest approach seems to be `{.strdefine.}` variables https://nim-lang.org/docs/manual.html#implementation-specific-pragmas-compileminustime-define-pragmas . For example 
```nim
const nbHomeDir {.strdefine.} = ""
when not defined(nbHomeDir):
  nbHomeDir = findNimbleDir(nbThisDir)
```
This won't run of course but it conveys the main idea. when compiling the file we could then pass in a directory like this `-d:nbHomeDir=McDonalds/BurgerKing` and if it isn't defined we use the nimble dir. 

Do you see any obvious flaws with this approach?

pietroppeter:
I do want to revise in general how paths are handled in nimib, since as we have abundantly seen in nimibook, current choices are not... great. Some kind of compile time switch on top of that to override whatever defaults we end up with could also be a good idea (I also like strdefine). Ideally I would like to revise everything first, but not to let the "better be the enemy of the good", if you come up with a working implementation for this I would proceed with a merge and release.

Ah, regarding nimiBoost, you should open up the issues, I have a suggestion/remark to do there. :)

HugoGranstrom:
> I do want to revise in general how paths are handled in nimib, since as we have abundantly seen in nimibook, current choices are not... great. Some kind of compile time switch on top of that to override whatever defaults we end up with could also be a good idea (I also like strdefine). Ideally I would like to revise everything first, but not to let the "better be the enemy of the good", if you come up with a working implementation for this I would proceed with a merge and release.

[...]

PR #61, opened 2021-06-24
nb refactor (fix #25) + new paths (fix #55) + fix #50, #51

### LINKED ISSUE #25: use a nb global object

### LINKED ISSUE #50: deprecate/remove nbPostInit
it did not look like a good idea from the beginning, but it proved to be really bad: https://github.com/pietroppeter/nimibook/issues/8
The alternative mechanism should be something like https://github.com/pietroppeter/nimib/pull/45 or https://github.com/pietroppeter/nimibook/pull/15

### LINKED ISSUE #55: [RFC] redesign nimib handling of paths
current status is mostly described here: https://github.com/pietroppeter/nimib#interaction-with-filesystem

current design has been challenged with the usage of nimib in [nimibook] and [nimiboost]

# proposal

## paths that remain the same:

* `nbThisFile` and `nbThisDir` remain `AbsoluteFile` and `AbsoluteDir` of current document source (where `nbInit` is called)
* `nbInitDir` remains `AbsoluteDir` of current directory (when `nbInit` is called)
* `nbDoc.filename` will remain a string containing an absolute or relative path and initialized as `nbThisFile`. It will be the reference from which `nbSave` will generate output filename (default behaviour is to change extension to html)

## paths/mechanism that will change:

* `nbHomeDir` will stay an `AbsoluteDir` but it will change its behaviour (no longer automatically initalized starting from Nimble file at the beginning). It is still interpreted as the root/base directory of the website, the one where 🏡 icon points to.
* a **new** `nbSrcDir` will be introduced (also an `AbsoluteDir`). It represent the directory in which `nim` source files are and the relative path of current source with respect to `nbSrcDir` should be replicated when producing a document in `nbHomeDir`
* **new** `nbRelUrl` proc that create a relative url string starting from a normal string (it can be a relative or absolute path). result will only contain forward slashes and it will contain the relative url from output document to the relative url with respect to `nbHomeDir` (basically adding the correct number of `../` as the `path_to_root` mechanism in [nimibook]).

## default behaviour

* both `nbHomeDir` and `nbSrcDir` are initialized as current directory. the behaviour is the same of a standard nim file with respect to writing or reading files (current directory is not changed) and html output will be next to nim file.

## compile time configuration of `nbHomeDir` and `nbSrcDir`

* with compile time define pragmas (i.e. `strdefine`) you can set `nbHomeDir` and `nbSrcDir`. The pragmas will be `nimibSrcDir` and `nimibHomeDir` (almost same as now, need to rename `nimibOutDir` to `nimibHomeDir`). They can be relative paths, and in that case they are made absolute with respect to current directory, or they can be absolute paths.
* if `nbHomeDir` is configured then current directory is also changed to `nbHomeDir` during `nbInit`
* if `nbSrcDir` is configured then `nbDoc.filename` is changed to a relative path with respect to `nbSrcDir`

## optional compile time anchoring mechanism

* the default anchoring mechanism with respect to nimble file is removed, a new and optional "anchoring" mechanism which generalizes this can be used.
* if `nimibAnchorPattern` is defined (through `strdefine`) then the system will find a `nbAnchorDir` (an AbsoluteDir), this will be used in place of current directory to initialize `nbHomeDir` and `nbSrcDir` if they are given as relative paths (no effect if they are given as absolute paths
* `nbAnchorDir` is found by starting from `nbThisDir`, checking if pattern is satisfied, if not move to parent directory
* pattern should support as before looking for a nimble file `*.nimble` but also looking for a specific file (e.g. `nbook.nim`) or a directory (e.g. `.git`)
* if `nimibAnchorPattern` is not provided or a pattern is not matched, `nbAnchorDir` is set to current directory

this anchoring mechanism should make document output independent of current directory (providing at the same time relative paths for `nimibHomeDir` and `nimibSrcDir`. It should also allow the possibility to compile single chapter of [nimibook] through `nim r`.

## to be decided

* should the change of directory during `nbSave` (`withDir(nbHomeDir): ...`) be removed? probably yes.

## notes

* this is breaking change and it will require fixes downstream in [nimibook] and [nimiBoost]

[nimibook]: https://github.com/pietroppeter/nimibook/
[nimiBoost]: https://github.com/HugoGranstrom/nimiBoost

pietroppeter:
fix #25 and #55 (also #50 and #51).

this will also contain breaking changes and next version including this will be 0.2.0 (whatever that means).

in particular I plan to not only deprecate, but directly remove the custom include mechanisms (fix #50, hopefully not used by anyone except me).

**edit**: most changes are done (see notes below). still to do:
- [X] new config based mechanism
- [x] prepare a PR for nimibook that will take advantage of these changes (https://github.com/pietroppeter/nimibook/pull/39)
- [x] a bit more logging of stuff (e.g. when saving mention if a directory is created? echo also name of file when saved?)
- [x] update documentation and changelog
- [x] bump to 0.2


pietroppeter:
notes:
* `nimibRootFindPattern` to be implemented
* added aliases in order to avoid most breakage (e.g. aliases for `nbDoc`, `nbBlock`, ...)
* added a `relPath` proc as before that uses `nb` object, but this is something that could be improved on with this approach.
* with current changes to run a Nimib example you can run `nim r docs\hello` but you cannot run anymore `nim r hello` while in `docs` folder (it will assume that `nb.homeDir=$projectRoot/docs/docs`); you can still run `nim r -d:nimibHomeDir=. hello` while in docs folder.

pietroppeter:
while working on a PR to have nimibook work with this I realized there is still one thing that I do not like about path handling and it is the fact that parameters `nimibSrcDir` and `nimibHomeDir` are compile time variables which make very little sense. This made me realize that those setting are **configuration** and made me rethink how they should be handled from first principles.

The two use case I have in mind:
* **single document**: a single nim file that generates an html. In this case I think the default would be that executable should not change the directory and that the html should be produced in the directory where the executable is run (this is different from previous default where html would be next to source; in fact binary could be even distribute and we cannot assume the source is even on the same filesystem; *what  happens currently in that case?*).  This would cover for example the case of an executable that runs a report based on data available in the directory in which it is run and generates the html file there.
* **project/website**: if a specific nimib **config file** is found in a parent directory this will be taken as root directory. in the config there can be a setting for a homeDir and srcDir (that would normally be relative to root dir). if a srcDir is present it is expected that source is present and lies somewhere inside srcDir and the output filename path is changed relative to this srcDir. If a homeDir is present then the directory is changed to be homeDir during nbInit.

I am thinking for the config file to use `toml` and in particular to use https://github.com/status-im/nim-toml-serialization to parse the config file. The `nb` object will have a `rawCfg: string` field, since this config could contain other configuration for themes (it could be accessed by nimibook theme).

I will try to materialize this thoughts in this PR and see how nimibook responds to it (my expectation is that nimibook codebase would be very happy to see this).

HugoGranstrom:
Oh... 🤯 I guess it's easy to fall into Nim's trap of doing everything as CT just because we can 😅 This seems nice though. Have to think a bit about how it affects NimiBoost, can we still pass in nbHomeDir and nbSrcDir at CT or is it only RT now?

pietroppeter:
currently (at most recent commit) we cannot pass homeDir and srcDir neither at CT nor at RT. The idea would be to make it available at RT, not sure if I manage in this PR. if not it will be in a later one. Regarding Nimiboost preview functionality, I would be interested also in investigating what are other options to obtain a similar functionality. The fact that nimiboost changes the output folder can make break the output html in many ways (if it needs images not found in that folder or other assets, like js, css, ..). I mentioned `nbShow` some time ago as an alternative to `nbSave`, I was thinking of making the option to "show in browser" a (runtime) option that would be available also when nbSave is called. I am also aware there are possibly a few "live html preview" extensions in VS Code (have not tried any one of them), those might be also useful.

HugoGranstrom:
OK 👍 You should definitely prioritize making Nimib's path handling good before we try and force NimiBoost into it 😄 The problem is that an external program must be able to influence runtime behavior and that's not an easy nut to crack. Would creating a temporary `config.toml` file in the same directory as the file in question work perhaps? If it is read during RT it feels like it could work 😅 We also have problems if there already is a toml file in that directory. In that case, it could be possible to add a `boost.toml` file that has higher priority, if we want to go this path.

> I was thinking of making the option to "show in browser" a (runtime) option that would be available also when nbSave is called. 

That is an option, it's always nice to not have to change your code too much 👍 It feels like the most natural approach, but perhaps not the most 

> I am also aware there are possibly a few "live html preview" extensions in VS Code (have not tried any one of them), those might be also useful.

So integrating Nimib with a VS Code extension directly? Or rather, instead of NimiBoost calling Nimib, we do it the other way around and let Nimib be the one who calls NimiBoost? Not sure if it is possible, but if it is...

HugoGranstrom:
Ugh, this was so confusing to read until I saw "edited by pietroppeter" 😅 

> a temporary nimib.toml file would be a workaround, I still think being able to give runtime options that override the config as a better solution.

I'm not having a great day understanding you today, sorry. What is a runtime option to you? A compile switch as we have now or something you have to add in the code?

> no, the idea would just be: there is this extension that show you a website in vs code with live preview, since nimib outputs html you can use it to preview nimib files (along with an extension with a shortcut to compile and run a nimib file). but as mentioned, I have not tried them. If on the site you have a html preview, on the other side you have a nim file and you have a command that compiles and run, you sort of have the feature, right?

Oh, you mean an extension that lets you view a HTML file from the file list manually? (for example: Right-click→"Preview HTML Document") 

The steps would be something like this then?
1. Compile nimib file (generates a HTML file)
2. Open HTML file with HTML Preview Extension
3. Update nimib file and recompile.
4. HTML preview is already open and updates automatically to new content

pietroppeter:
Ops sorry I think I hit Edit instead of "quote reply" 😬

Runtime option would be something I need to add to the code. `nim c example`, then `./example --nbHomeDir=xxx`...

Yes, that was my thinking about Html preview.

pietroppeter:
ok, I reverted my edit that deleted part of your comments. my reply is not there but is reported in your reply to my edit, so I guess I will leave it like that without trying to do more damage...

and just to try and avoid more misunderstanding (something which I am not very good...), when I reply that for the runtime option I need to add code, I mean that I need to add it to the library, not the user will need to add it himself.

HugoGranstrom:
No worries, it was just a bit confusing seeing words I didn't remember myself writing 🤣


Oh, so there isn't much changing then for NimiBoost actually? Now we do `nim r -d:nimibHomeDir=some_path file.nim`, but with your new runtime options we just do `nim r file.nim --nimibHomeDir=some_path`?

> Yes, that was my thinking about Html preview

Ok, then I understand 😄 But if my suggestions above works, the old way of doing it should still work right? Then the user can choose which method they want to use. 

> and just to try and avoid more misunderstanding (something which I am not very good...), when I reply that for the runtime option I need to add code, I mean that I need to add it to the library, not the user will need to add it himself.

Thanks for explaining, I think we are on the same page now 😁 (who knew RT vs CT could be this multi-faceted🙈)

pietroppeter:
Yes, I think we are on the same page now! 😄

pietroppeter:
having added the runtime options now this also fixes #51

PR #63, opened 2021-07-13
WIP nbCodeBlock macro

### LINKED ISSUE #20: Preserve code formatting in a block same as in source code
Hello! I just started using nimib, and I did see that the website says
> nbBlock.code: stringification of the code block through AST. if it appears different than what you typed it is because nim likes it better that way. In particular only documentation comments survive this process and normal comments will not appear.

Would it be possible to allow nbCode to preserve the original spacing/formatting? My currently built HTML page looks like https://ajusa.github.io/binarylang-fun/intro.html, and the newlines after each colon look pretty strange. The original code is here: https://github.com/ajusa/binarylang-fun/blob/master/intro.nim. Is there a flag or something to make sure that the original spacing is preserved?

EDIT: Another issue I noticed is that curly braces in the code seem to have vanished in the code output? Is this related to a mustache template somehow? Apologies if I missed this in the docs somewhere.
Input:
```nim
s: _ = "\n"
*header: {headers}
s: _ = "\n"
```
Output:
```nim
  s:
    _ = "\n"
  *header
  s:
    _ = "\n"
```

HugoGranstrom:
Fixes #20 

This is a mix of @haxscramper and @Yardanico codes which reads the source file directly to circumvent the (sometimes ugly) formatting from the compiler. 

I haven't done any extensive testing yet but it seems to work ok for `numerical` at least. ~~Have to do something about the `mixin`s as well. It could be solved by doing all the string operations at CT, but do we want that?~~

Rough TODO:
- [x] Handle the single-line case `nbCode: echo "Hello World"`
- [x] Handle the case when we have comments at the end of the code block. They aren't accounted for in the lineInfo. 
- [x] Handle the case when we end on a triple quoted string. 
- [x] Investigate if it can be put into macros or templates without hacks (eg codeInBlock)


HugoGranstrom:
A problem I've stumbled upon regarding trailing comments:

```nim
nbCode:
  echo "Hello World!"
  # This comment gets included
  # This one as well
  
  # But not this one 
```

In VSCode at least, by default, it doesn't add any indentation on the empty line so if I just check the indentation it won't recognize it as part of the block as it will break on that line. So I'm thinking that one could add empty lines as well to the block. I'm not sure if there are any cases where it would cause the code to expand further than the block and end up eating the whole file 🤔

pietroppeter:
Hi @HugoGranstrom, I am currently on vacation with often bad internet, not sure when I will review this but thanks for working on it.

HugoGranstrom:
@pietroppeter No hurry, I think I have things left to tinker with still 😄 And sorry in advance for messing up your `ptest` 😅 I'm trying to go through as many Nimib files as possible and manually checking they produce the correct code. But I bet there are some edge cases my eyes have missed so when you eventually take a look at it, just try your best at tricking it 😉

pietroppeter:
hey, I have not forgotten about this and I also had a little time to think about, here are some thoughts:
* this functionality should be available behind a compile time switch `nimibPreviewCodeAsInSource`. Later on it will be come the default functionality but if feasible I would like to keep old mechanism available behind a switch.
* I think `nbInit` should make available a `const nbSource` at compile time with the source
* easier composability of this would be much useful. would it be possible to have a function with something like this signature:

```nim
macro getCodeAsInSource(source: static string, command: static string, body: untyped): static string
```

* the above could be called in place of current call to `toStr` (maybe changing how it is used). the `command` argument would be `nbCode` in the case of `nbCode` template but it will be other commands in other cases (I guess it is necessary to extract the correct piece of source).
* if I understand correctly the `startPos` and `endPos` functions give start and end lines that have to be corrected using the source and parsing indentation, right? can we extract some normal function (more easily testable) that is a normal function with a signature like:

```nim
func getCodeBlock(source: string, command: string, startPos: Pos, endPos: Pos): string
```
* I guess the above could be called by the above macro but it could be easier to test.
* I plan to remove `ptest` mechanism and replace it with some proper testing, so do not worry about that.

HugoGranstrom:
You bring up a lot of good points 😄 A switch from old to new parser sounds good, and yes composability isn't really great atm. 

>  the command argument would be nbCode in the case of nbCode template but it will be other commands in other cases (I guess it is necessary to extract the correct piece of source).

What other cases for `command` do you have in mind? I can't come up with any other block than `nbCode` which would want to use `getCodeAsInSource` 🤔 

> if I understand correctly the startPos and endPos functions give start and end lines that have to be corrected using the source and parsing indentation, right? can we extract some normal function (more easily testable) that is a normal function with a signature like:

Indeed 👍 Should be perfectly doable. The macro-mess I've created would be better off with a refactoring. Having the code file come in as a string would significantly make things easier. 

> I plan to remove ptest mechanism and replace it with some proper testing, so do not worry about that.

👍 

> the above functions could be also in their own module if it makes sense to separate them (name sources?)

Sounds good to me! `sources` or `parsers`, depending on what more you would like to add here in the future.

pietroppeter:
> > the command argument would be nbCode in the case of nbCode template but it will be other commands in other cases (I guess it is necessary to extract the correct piece of source).
> 
> What other cases for `command` do you have in mind? I can't come up with any other block than `nbCode` which would want to use `getCodeAsInSource` 🤔

e.g. `nbFile` (something that saves content in a nim file). there are others mentioned in other issues.

> > the above functions could be also in their own module if it makes sense to separate them (name sources?)
> 
> Sounds good to me! `sources` or `parsers`, depending on what more you would like to add here in the future.

I would go with `sources`, I really do not see it as a parser at the moment (although I guess it could be thought as one).

pietroppeter:
> What would source be in the case of nbCode then? Just the current file? Or do we just ignore it in that case?

not sure how this relates to alternatives to `nbCode`, in any case the `source` would be the static source of current file that would need to be created as global (a `const nbSource`). But this is just to avoid reading again and again the source of current file in every nbCode (or other) template/macro.

HugoGranstrom:
Then I'm confused as `command` is all about alternatives to `nbCode`? I'm not really seeing what purpose it would have in other cases where it would matter how we read the source. This is all that's needed in my head:
``macro getCodeAsInSource(source: static string, body: untyped): static string``
Nim code is Nim code regardless of if it's going to be saved in a file (`nbFile`) or just shown `nbCode`. What am I missing? 😅

HugoGranstrom:
> I am assuming that nbcode hard-coded string in your code is used to more precisely pinpoint the lines of code to extract form the source. If that is the case, a different command should require a different string I guess.

Ohhhhhhhhhhhhhhhhhh 🙈 Now I see what you mean. You are referring to this proc, right?

```
proc isNbCodeLine*(s: string): bool =
  "nbcode" in s.toLower
```

And in the case of `nbFile` we want `"nbfile" in s.lower` instead?

HugoGranstrom:
I have obviously already forgotten how my code looked 🤣

pietroppeter:
yep, that was it! (in principle it should also remove underscores for those who will write `nb_code` and it should be really identifier equality i.e. be case sensitive on first character...)

I forget all the time my code, that's why I write it down :D

HugoGranstrom:
> yep, that was it! (in principle it should also remove underscores for those who will write nb_code and it should be really identifier equality i.e. be case sensitive on first character...)

It feels like there must be a built-in function for that in the compiler somewhere... And yes there is: https://nim-lang.org/docs/strutils.html#nimIdentNormalize%2Cstring . There is still the problem of isolating it from the rest of the line though :/ For example `nb_FiLe("main.nim"):`. Or perhaps it just works. Yeah it might actually work :O

> I forget all the time my code, that's why I write it down :D

XD you write down your code?! What a revolutionary idea!

HugoGranstrom:
TODO for refactor:

- [ ] Add `const nbSource` to `nbInit` that reads file at compiletime. 
- [ ] Move functions to `sources.nim`
- [ ] Add commandline switch `nimibPreviewCodeAsInSource` to switch on the new parser.
- [ ] Write `getCodeAsInSource` macro that takes `command` as input.
- [ ] Write general `getCodeBlock` function which take the raw `startPos` and `endPos` and extract the code block with some additional logic. 

May add more if I find I missed something

HugoGranstrom:
@pietroppeter I threw this together in a rush but I'm able to get the `numerical` file to compile and it seems fine. So it should be ready for a test run. Have I missed anything from your specifications? 😄

pietroppeter:
from a quick look it seems great, thanks! I want to check it out and test it myself but it looks very nice!

HugoGranstrom:
Nice! Shout if you find any bugs and I'll try and fix them as soon as possible in that case

src/nimib.nim
```diff
@@ -64,13 +66,30 @@ template nbInit*() =
...
-    nbCodeBlock(nbBlock, nbDoc, body)
+    nbCodeBlock(nbSource, nbBlock, nbDoc, body)
 
-  template nbCodeInBlock(body: untyped) =
+  template nbCodeInBlock*(body: untyped): untyped =
     block:
       nbCode:
         body
 
+#[
```
pietroppeter, line 76: this (and other similar commented code) can probably be removed when everything is tested, correct?

src/nimib/sources.nim
```diff
@@ -0,0 +1,202 @@
...
+import std/[
+  parseutils, 
+  strutils
+  ]
+import types
+
+
+
+# Credits to @haxscramper for sharing his code on reading the line info
+# And credits to @Yardanico for making a previous attempt which I have taken much inspiration from.
```
pietroppeter, line 11: you could specify that I means you :)

src/nimib/sources.nim
```diff
@@ -0,0 +1,202 @@
...
+type
+  Pos* = object
+    line*: int
+    column*: int
+
+proc toPos*(info: LineInfo): Pos =
+  Pos(line: info.line, column: info.column)
+
+proc startPos*(node: NimNode): Pos =
+  ## Has column info
```
pietroppeter, line 22: this (and similar in finishPos) look more like normal comment than function documentation, correct?

src/nimib/sources.nim
```diff
@@ -0,0 +1,202 @@
...
+  else: # single line case, eg `nbCode: echo "Hello World"`
+    let line = lines[startLine]
+    codeText = line.split(":")[1 .. ^1].join(":").strip() # split at first ":" and take the rest as code and then strip it.
+  return codeText
+
+macro getCodeAsInSource*(source: static string, command: static string, body: untyped): static string =
+  ## substitute for `toStr` in blocks.nim
+  let startPos = startPos(body)
+  let endPos = finishPos(body)
+  let lObj = body.lineInfoObj()
```
pietroppeter, line 121: this lObj can probably be removed now

pietroppeter (changes requested):
I like the code and now it can even support the templates, so that is great. just a few minor changes but I will add a bug that I was able to find in the PR conversation

pietroppeter:
looks very nice overall, I found a bug that I guess could be fixed with a startPos. You could probably add a tsources .nim in tests (wrapped in unittest or similar) with something like this (second case shows the bug):

```nim
import nimib, strutils, strformat

template check =
  echo &"---\n{nbBlock.code}\n"
  # the replace stuff needed on windows where the lines read from file will have windows native new lines
  assert nbBlock.code.replace("\c\l", "\n").replace("\c", "\n") == expected, &"---ERROR:\n{nbBlock.code.repr}\n---\n{expected.repr}\n---"

var expected: string
nbInit
nbCode:
  # a comment
  let
    x = 1
# a comment here means that on windows line 12 will read with a single \c instead of \c\l (if blank space here I will have \c\l)
expected = """
# a comment
let
  x = 1
"""
check

nbCode:
  # a comment
  let  # and a comment with nbCode
    y = 1
expected = """
# a comment
let  # and a comment with nbCode
  y = 1
"""
check
```

src/nimib/sources.nim
HugoGranstrom: Yep, it's still just a copy paste of hax's code so will go through and maybe add a few documentation comments as well.

src/nimib/sources.nim
HugoGranstrom: Yes, was only needed for debugging purposes

HugoGranstrom:
Thanks for the bug hunt :D I'll look into them. If I understand your code samples correctly, these are the case:
- Something related to newlines
- single line `nbCode`s without `:`

Is that correct?

I'll try and get some stuff done today, but should be finished at the end of the weekend as latest if it isn't a rabbit hole to fix stuff :)

HugoGranstrom:
ok I see the problem now, it mentions `nbCode` in the comment and interprets it as the start of the code block...

pietroppeter:
To be honest in all of this I am still puzzled how we cannot get the correct line numbers from Nim. Is this something that could be considered a bug and we should submit an issue?

HugoGranstrom:
> To be honest in all of this I am still puzzled how we cannot get the correct line numbers from Nim. Is this something that could be considered a bug and we should submit an issue?

If I read the code correctly it is two main cases which we cover:
1. Ordinary `#` comments
2. The compiler only give us the *beginning* of the last statement, so for multiline expressions or multiline strings we must find their ends. Idk if this is something the compiler should do or not.

HugoGranstrom:
A question: How many weird ways of writing code do we want to support? When digging deeper and deeper into this it feels like there will inevitably be edge-cases for which it doesn't work for. Case-insensitivity in addition to whitespace-insensitivity (`nbCode(echo "Hello")` is same as `nbCode      (echo "Hello")    `) makes it hard to keep the code concise. If it had just been one of them it would've been quite straight forward, but extracting the code from for example:
```nim
nb_Code    (echo "Hello")
```
is really hard as we can't just normalize the code because then indexes will change (`nbCode` is shorter than `nb_Code`) or the outputted code will be normalized. So in such cases, one could as well just use `toStr` directly. 

This is mainly concerning single-line blocks though, but I'm sure multiline blocks have their problems too. So how robust do you want this to be? Should it be an opt-in tool that you have to write "reasonably idiomatic" Nim code to use or do you want it to cover all valid Nim code?

pietroppeter:
If we find a way to manage many weird ways in a sane way we should do it. If not we will be pragmatic and require idiomatic formatting for this feature to work.

On this specific weird ways a possible strategy could be:
- get correct first line (for this we can normalize line code, strip white space and make sure it starts with our command)
- get correct last line (processing multi line strings and whatever else)
- once we have done that we now for sure that our first line begins (except for white space) with some weird variation of our command, we process the original line to find next chat after our command. It can be a `:`, a `(`, or a whitespace and from there we should be able to figure out the rest of the content, hopefully

Not sure if the above plans cover all issues and I was a bit sketchy on last step but it does not sound impossible to get right.

Anyway I think a first version more or less like it is now is already useful, so we could wrap it up and leave it for the next lucky guy who wants to break his head on this :)

HugoGranstrom:
> If we find a way to manage many weird ways in a sane way we should do it. If not we will be pragmatic and require idiomatic formatting for this feature to work.

Yes if someone comes up with such a solution it's the best solution. I haven't been able to.

> On this specific weird ways a possible strategy could be:
> - get correct first line (for this we can normalize line code, strip white space and make sure it starts with our command)
> - get correct last line (processing multi line strings and whatever else)

These steps I'm on board with 👍 it's basically what we do right now. The only problematic thing is if the first line doesn't start with `nbCode` (ie `block: nbCode: echo "Hello"`), then we have the same problem as in step 3.

> - once we have done that we now for sure that our first line begins (except for white space) with some weird variation of our command, we process the original line to find next chat after our command. It can be a :, a (, or a whitespace and from there we should be able to figure out the rest of the content, hopefully

> Not sure if the above plans cover all issues and I was a bit sketchy on last step but it does not sound impossible to get right.

It is here that the fun begins 😅 how do we know where the strange variation of `nbCode` ends? We normalize the line. Now how do we know where it ended in the original string? This is the main problem I'm fighting against. We can't just go between these two strings and extract indices from one by only looking at the other. This is the reason I hardcoded `command = nbCode` to begin with so we only had to worry about whitespace. I'm lost right now on how to do this without writing my own normalizing parser 🤷 So if you have any insights I'm all ears 😉 

> Anyway I think a first version more or less like it is now is already useful, so we could wrap it up and leave it for the next lucky guy who wants to break his head on this :)

Haha, poor soul whoever that turns out to be 🤭 It's just the single line stuff that's left before I'm content with calling this done :)

pietroppeter:
Well we could post on the forum or on stack overflow, I think someone (for example @shirleyquirk) might appreciate the challenge 😉

HugoGranstrom:
Remember this? 😅 
```nim
proc startPos*(node: NimNode): Pos =
  ## Has column info
```
> **Has column info**

🤦 

We have the start of the expression already... 🙃 This handles the `block: nbCode: echo "Hello"` problem, now it's just a matter of making it work for the multiline case as well:
```nim
block: nbCode:
	echo "Hello"
```
And in this case, we only care about whether it is the start line or not, no index magic needed. Problem with relaxing the "nbCode should be first thing on line" condition is that it would be triggered by `nbCode` being in strings and comments:
```nim
nbCode:
  echo "nbCode" # nbCode
```
Both of these are problematic.

HugoGranstrom:
But if you are happy with that limitation for now (requiring `nbCode` to be the first thing on the line), I'm happy as well 😛

src/nimib/sources.nim
```diff
@@ -101,20 +100,30 @@ func getCodeBlock*(source: string, command: string, startPos, endPos: Pos): stri
...
+    var extractedLine = line[startPos.column .. ^1].strip()
+    if extractedLine.strip().endsWith(")"):
+      # check if the ending ")" has a matching "(", otherwise remove it.
+      var nOpen, nClose: int
+      for c in extractedLine:
+        if c == '(': nOpen += 1
+        elif c == ')': nClose += 1
+      if nOpen < nClose:
+        # remove trailing ")" so that we get matching number of "()".
+        # We trust the compiler to only give us valid code, thus we can assume all ")" are at the end as the missing "("
```
shirleyquirk, line 122: if you're parsing code char by char, to handle edge cases you end up needing more of a parser, i'm afraid,
you end up tripping over parens inside of quotes, and then tripping over escaped quotes.
[here](https://github.com/nim-lang/Nim/blob/a78ee8ae8453aabde04fa0103e02378efe870827/lib/pure/strformat.nim#L601-L628) is what i had to do in strformat

HugoGranstrom: Ugh, you are right, nice catch 😄🙈 

That looks like quite a lot of work to find all edge cases 😅 What do you think about just counting the opening brackets before `startPos.column` until we hit a non-whitespace character that isn't an open parentheses? Or are there any other characters which are valid to put where I mark in this code?

```nim
nbCode ( echo "y" )
#            ↑ ↑
```

pietroppeter:
> But if you are happy with that limitation for now (requiring `nbCode` to be the first thing on the line), I'm happy as well 😛

I think that is a fine limitation for now.

If you want to wrap it up we can already merge this which is a fine improvement.

Not sure what is left to do:
- the "Investigate if it can be put into macros or templates without hacks (eg codeInBlock)" has been done for some time right? we could check the box in top comment
- [ ] test file is there, I think we miss a command in CI to run it (a `nim r -d:nimibPreviewCodeAsInSource tests\tsources`?)
- [ ] write a few lines that summarize the feature, the limitation (where it works) and some know cases where it does not work.
- whatever clean up you feel like doing

On this last point, I am not sure where to put that. It could be just some text here in the PR and then I copy in the release notes for 0.2.4 (I have not updated changelog after 0.2, I was planning to do that when releasing 0.3).

pietroppeter:
and thanks @shirleyquirk for helping out reviewing this, sorry for summoning you up like this 😊

HugoGranstrom:
Sounds like a plan. I'll try and fix @shirleyquirk's point, but then I'm fine for good 😁 When that's done I'll add a test command to run the tests and write a few lines here for the release 👍

HugoGranstrom:
I think I'm done now, added a test command `nimble test` that you can add whatever other tests you add in the future if you'd like. Here's a short text describing the feature:

Using `-d:nimibPreviewCodeAsInSource` Nimib will read the source file directly and get the code from it instead of letting the Nim compiler render a cleaned-up version of the code. This means formatting and ordinary comments (`#`) will be preserved (no more secrets!). There are some limitations though. It isn't 100% accurate yet so if you find that it doesn't show the code you expected, please open an issue. The known issues right now are:
- It doesn't work if the `nbCode` isn't first on the line. For example this doesn't work:
```nim
# This doesn't work!
block: nbCode: 
	echo "Hello World"

# This does work!
block:
	nbCode:
		echo "Hello World"
```
- If the last expression of the code block is a multiline expression, there might be edge-cases we haven't accounted for yet. But it should mostly work. 

Any feedback on it? :)

pietroppeter:
Ah, I see this only now... looks good, thanks merging now!

pietroppeter (approved):

pietroppeter:
Mmh there is still a conflict, can you resolve that?

pietroppeter:
conflict solved (yesterday on mobile I was not able)

pietroppeter:
ok, now there is a failure in the test you added :), not sure what the problem is though 🤷‍♂️

HugoGranstrom:
Hehe I was a bit lazy, the Github interface didn't let me resolve it for some reason 🤔

> ok, now there is a failure in the test you added :), not sure what the problem is though 🤷‍♂️

Huh, I pulled the merge commit and it runs fine both on Windows and WSL for me 🤥 What error do you get?

pietroppeter:
the error is in the CI (which I did not realize never ran because I did not approve anything...), on this line and subsequent one: https://github.com/pietroppeter/nimib/runs/4221322003?check_suite_focus=true#step:7:119

HugoGranstrom:
Thanks, I'll take a look when I get home later today. :) But it looks like it's perhaps only the testing itself which is flawed, some whitespace-thingy. Strange that it worked offline though. Have you tried running it locally?

pietroppeter:
Running locally on Windows works fine, have not yet tested with WSL (installing `gh` fails...). Looking at the error it seems like the actual multiline has no new line at the end, which I cannot reproduce locally but I guess it can be tied to capturing stdout in a wrong way.

tests/tsources.nim
```diff
@@ -0,0 +1,49 @@
+import nimib, strutils, strformat
+import unittest
+
+suite "test sources":
+  template check =
+    #echo &"---\n{nbBlock.code}\n"
+    # the replace stuff needed on windows where the lines read from file will have windows native new lines
+    test $currentTest:
+      check nbBlock.code.replace("\c\l", "\n").replace("\c", "\n") == expected
```
pietroppeter, line 9:
```suggestion
      actual = nbBlock.code.replace("\c\l", "\n").replace("\c", "\n")
      echo &"===\n---actual:\n{actual.repr}\n---expected\n{expected.repr}\n---\n==="
      check nbBlock.code.replace("\c\l", "\n").replace("\c", "\n") == expected
```

tests/tsources.nim
```diff
@@ -0,0 +1,49 @@
...
+    currentTest += 1
+
+  var currentTest: int
+  var expected: string
```
pietroppeter, line 13:
```suggestion
  var expected, actual: string
```

pietroppeter:
I am trying to add some echos to confirm that

HugoGranstrom:
Yeah, seems like it. Locally I get the trailing newline in all cases. Is there something we can do about this or is it just CI-dependent? In the worst case we just have to write a separate test for the CI without trailing end-lines 😅

pietroppeter:
anyway this failing check is not an issue concerning this PR, so I am merging this (and later will release).

It seems to me the issue now is in the fact that `captureStdout` works differently on CI then locally on linux. I will add a test for that and I will try to figure out how to make it work.

To wrap up on this specific PR, thanks for the work and perseverance on this @HugoGranstrom!

HugoGranstrom:
Ok 👍 that sounds like a beast one doesn't want to spend hours fighting 😱 

My pleasure 😄

pietroppeter:
I realize now that in the end we do not actually need to read the source statically (we just derive startPos and endPos statically and find the correct positions during runtime). makes it for a simpler api, nice!

HugoGranstrom:
Yes, changed it when porting to nimib 2.0 and it was nice not having to rely on `static string` everywhere :)

pietroppeter:
fixed tsources here: https://github.com/pietroppeter/nimib/commit/9776ee0125ee9ee60af3671654597edb17f37c7d

PR #64, opened 2021-11-05
add nbFile

Clonkk:
Implement an elementary version of nbFile https://github.com/pietroppeter/nimib/issues/34


Clonkk:
@pietroppeter CI failure unrelated ^^'

pietroppeter:
Yeah I guess, also on todo is to remove ptest approach and add some regular testing

pietroppeter:
The block as is I guess is a fine and simple way to write stuff to file. The rendering I guess it is a bit more tricky. You could go with adding two blocks: one of text with "`file.Nim`:". And a code block with body content (without executing).

Clonkk:
>  one of text with "file:

Sure, that's a good idea.

> And a code block with body content (without executing).

What if the content of the file isn't nim code ? Typically, I'd like to use it for writing python / Julia stuff for https://github.com/SciNim/getting-started/pull/34

pietroppeter:
Yep, that was not the original use case but it looks a useful one. Could it work to have a `nbFile` overload that takes a string as body argument?

pietroppeter:
The only problem now would be that a code kind will receive a `lang-nim` class and it would be highlighted wrongly :/

Clonkk:
> Yep, that was not the original use case but it looks a useful one. Could it work to have a nbFile overload that takes a string as body argument?

I guess we could have both. The string version for generic type that just writes the file and the untyped version that interprets it as Nim code, that should work.

Clonkk:
Also, I think we should have different font for the code output generalyl speaking. Atm, it looks like text and can make it hard to distinguish between paragraph and code output.

pietroppeter:
agree, I created this issue to discuss improving code output rendering: #65

Clonkk:
I added some stuff (i'll finish later) but you can still take a look for now at the idea :).

Also, do you know why adding "> " to identBlock.output lines doesn't turn into a quote ?

pietroppeter:
Let's not change the output formatting in this PR please

Clonkk:
Oh OK, will revert

pietroppeter:
> Also, do you know why adding "> " to identBlock.output lines doesn't turn into a quote ? I was under the impression that renderHtmlOutput called renderMarkdown 

No, it does not call renderMarkdown, it wraps output in html directly with renderCodeBlockOutput

Clonkk:
@pietroppeter I reversed all the change regarding output formatting and just kept nbFile

pietroppeter:
added a few comments. could you also add an example or screenshot to see how this would look like? thanks!

Clonkk:
> added a few comments

Hm, I don't see comments on the PR ?

> could you also add an example or screenshot to see how this would look like? thanks!

So far, the rendering is not super pretty because I have to use  'nbText' so the formatting gets lost on display; I think a follow up PR regarding the formatting output is necessary.

pietroppeter:
> > added a few comments
> 
> Hm, I don't see any comments on the PR ? Is that GH bug :D ?
> 

they should be available as a review: here is how I see them

image


> > could you also add an example or screenshot to see how this would look like? thanks!
> 
> So far, the rendering is not super pretty because I have to use 'nbText' so the formatting gets lost on display; I think a follow up PR regarding the formatting output is necessary.

two of the comments above are suggestions about formatting and I would expect they provide a minimally decent formatting (e.g. wrapping the code in triple backticks).

Clonkk:
@pietroppeter Fixed the comments; I added a small thing when when writing file I sue the extension of the file as the markdown rendering indication so 

a.json ->  ```hjson
a.py -> ```py 

etc...

pietroppeter:
looks good to me, thanks for this!

adding the extensions is also good for me, maybe later we could improve by adding the full name of language for some known extensions so that a tool like highlight.js can automatically highlight (see https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md, python already supports py as alias, but julia does not support jl).

If there is nothing else you want to add I can merge and later I will cut a new release (0.2.2)

Clonkk:
> adding the extensions is also good for me, maybe later we could improve by adding the full name of language for some known extensions so that a tool like highlight.js can automatically highlight (see https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md, python already supports py as alias, but julia does not support jl).

Yeah I noticed that; eventually that's something we can fix by just having a table or extension -> highlight for language but there's because "```randomstuff " is still valid markdown (it just doesn't high light anything).

> If there is nothing else you want to add I can merge and later I will cut a new release (0.2.2)

For this PR I think it's good. We can work on https://github.com/pietroppeter/nimib/issues/65 on release 0.2.3

PR #70, opened 2021-11-13
Update penguins example, closes #69

### LINKED ISSUE #69: update penguins example to use datamancer

Vindaar:
This updates the penguin example with the following changes:

- use datamancer and ggplotnim separately for clarity
- classification by shapes is now better supported, use it
- add fitted lines to Simpson's paradox plot

In addition I added a comment explaining equivalents to the (more or less deprecated `c""` notation).


docs/penguins.nim
```diff
@@ -52,34 +54,34 @@ nbImage(url="images/penguins_map.png",
...
-To plot this we need to remove all `NA`s:
+To plot this we need to remove all `NA` and then classify the points both by the penguins sex as
+well as their species:
 """
 # manage runtime error here!
 #nbCode:
-#  ggplot(df, aes(x="body_mass_g", y="flipper_length_mm", color = "sex", shape="species")) + geom_point() + ggsave("images/penguins_mass_vs_length_with_sex_and_species.png")
-#nbImage(url="images/penguins_mass_vs_length_with_sex_and_species.png", caption="Penguins' mass vs flipper length")
+#  ggplot(df, aes(x="body_mass_g", y="flipper_length_mm", color = "sex", shape="species")) + geom_point() + ggsave("images/penguins_mass_vs_length_with_sex_and_species2.png")
+#nbImage(url="images/penguins_mass_vs_length_with_sex_and_species2.png", caption="Penguins' mass vs flipper length")
```
pietroppeter, line 63: we probably can remove this whole commented block now that shape works :)

docs/penguins.nim
```diff
@@ -52,34 +54,34 @@ nbImage(url="images/penguins_map.png",
...
 nbCode:
-  let df1 = df.filter(f{c"body_mass_g" != "NA"}) # why c""?
-  ggplot(df1, aes(x="body_mass_g", y="flipper_length_mm", color = "sex")) + geom_point() + ggsave("images/penguins_mass_vs_length_with_sex.png")
+  let df1 = df.filter(f{`body_mass_g` != "NA"}) # c"foo" == `foo` == idx("foo") (accent quoted not usable for columns w/ spaces)
```
pietroppeter, line 65:
```suggestion
  let df1 = df.filter(f{`body_mass_g` != "NA"}) # c"foo" == `foo` == idx("foo") (backtick quotes not usable for columns with spaces)
```

out of curiosity: really backtick quotes are not working for columns with spaces? seems odd knowing that they can be used for variable names with spaces, but I imagine there are good reasons for that

docs/penguins.nim
```diff
@@ -52,34 +54,34 @@ nbImage(url="images/penguins_map.png",
...
 - males are in general bigger than females but there appear 2 groups, possibly related to species
-- we have some more NAs (and one '.') in sex column (even after filtering for NAs in numeric columns)
-
-Let's now look at the same chart coloring by species instead of by sex (shape is not yet supported in ggplotnim):
-"""
-nbCode:
-  ggplot(df1, aes(x="body_mass_g", y="flipper_length_mm", color = "species")) + geom_point() + ggsave("images/penguins_mass_vs_length_with_species.png")
-nbImage(url="images/penguins_mass_vs_length_with_species.png", caption="Penguins' mass vs flipper length (colored by species)")
-nbText: """We can see that sizes of Adelie and Chinstrap overlap, while Gentoo penguins are in general bigger.
+- we have some more NAs (and one '.') in sex column (even after filtering for NAs in numeric columns
```
pietroppeter, line 72:
```suggestion
- we have some more NAs (and one '.') in sex column (even after filtering for NAs in numeric columns)
```

docs/penguins.nim
```diff
@@ -52,34 +54,34 @@ nbImage(url="images/penguins_map.png",
...
+As a final plot, I would like to (partly, confidence bands are missing) reproduce a plot that "shows" the presence of [Simpson's paradox](https://en.wikipedia.org/wiki/Simpson%27s_paradox) in this dataset,
 as reported by [this tweet](https://twitter.com/andrewheiss/status/1301166792627421186):
 """
 nbCode:
-  ggplot(df1, aes(x="culmen_depth_mm", y="body_mass_g", color = "species")) + geom_point() + ggsave("images/penguins_simpson.png")
+  ggplot(df1, aes(x="culmen_depth_mm", y="body_mass_g")) +
+    geom_point(aes = aes(color = "species")) +      # point...
+    geom_smooth(aes = aes(color = "species"),       # and smooth classified by species
+                smoother = "poly", polyOrder = 1) + # polynomial order 1 == line
+    geom_smooth(smoother = "poly", polyOrder = 1) + # and smooth w/o classification by species
```
pietroppeter, line 83:
```suggestion
    geom_smooth(smoother = "poly", polyOrder = 1) + # and smooth without classification by species
```
as above, I always prefer the full with/without expression, better for those not familiar with the abbreviation

pietroppeter (changes requested):
a few minor points, thanks for this!

docs/penguins.nim
Vindaar: Ah, yes. I wasn't sure about that, because that code still tries to plot with the NAs in the DF. And that still doesn't work (I don't actually know why, need to check what's going on). Will remove it and track it in a ggplotnim issue.

docs/penguins.nim
Vindaar: Yes, it's because of a simple limitation of the Nim AST and how it treats these quotes. 

The problem is the following: Anything in accented quotes (as they are called in the Nim AST), will be added to a "list" of identifier parts. So 
```nim
dumpTree:
  `hello`
```
is 
```nim
  AccQuoted
    Ident "hello"
```
If there's a space in there, it becomes:
```nim
dumpTree:
  `hello world`
  AccQuoted
    Ident "hello"
    Ident "world"
```
Seems fine so far right? We just concat the multiple idents using spaces! Look closer though:
```nim
dumpTree:
  `Energy[keV]`
  `Energy [keV]`
```
which gives us:
```nim
  AccQuoted
    Ident "Energy"
    Ident "["
    Ident "keV"
    Ident "]"
  AccQuoted
    Ident "Energy"
    Ident "["
    Ident "keV"
    Ident "]"
```
see the difference? Exactly, there isn't one. So we lose information about where spaces are in certain contexts (i.e. things between which there are already multiple separate identifiers). 

One could choose one or the other (always space, never space) or do some heuristics, but in the end I simply tell people to not use it for spaces or anything that is more complicated. Think about it for columns made up of `{'a' .. 'Z', '0' .. '9'}` and you're good. The `c""` notation was my first attempt to allow for a fallback. Since rewriting a big chunk in datamancer I decided though to go for a more specific `idx("foo")` approach for general string based things (`idx` can also take a string variable for instance. Main reason was to also provide a distinction between referring to a single element of a column (via `idx` which we loop over) or a *full column* using `col("foo")`.

pietroppeter:
All solved, right? The only thing left I see is to generate again the html (which is a bother, I want to pass to a CI more similar to scinim getting started)

Vindaar:
> All solved, right? The only thing left I see is to generate again the html (which is a bother, I want to pass to a CI more similar to scinim getting started)

You're right I forgot to update the HTML. 
Indeed having an auto generate + deploy preview would be great. @HugoGranstrom ? :hand_over_mouth:

HugoGranstrom:
I have been summoned! 

Depending on if you only want an automatic deploy on merge to Github Pages or if you want a Preview á la Getting Started for PRs it should just be to copy some stuff from the Getting Started CI 😄 Let me know if you want me to fix anything @pietroppeter

pietroppeter:
I have expressed my wishes in #71 (we should continue the conversation there).

PR #80, opened 2022-03-10
A bit of everything related to v0.3

HugoGranstrom:
Creating this now so I don't forget some parts. 
TODO:
- [x] Missing `mustache` import
- [x] nbClearOutput
- [x] nbRawOutput
- [x] add `readCode` parameter to `nbNewBlock`
- [x] Write some tests

### DIFF
--- a/src/nimib/blocks.nim
+++ b/src/nimib/blocks.nim
@@ -16,15 +16,16 @@ func nbNormalize*(text: string): string =
   text.replace("\c\l", "\n").replace("\c", "\n").strip # this could be made more efficient
 # note that: '\c' == '\r' and '\l' == '\n'
 
-template newNbBlock*(cmd: string, nbDoc, nbBlock, body, blockImpl: untyped) =
+template newNbBlock*(cmd: string, nbDoc, nbBlock, readCode, body, blockImpl: untyped) =
   stdout.write "[nimib] ", nbDoc.blocks.len, " ", cmd, ": "
   nbBlock = NbBlock(command: cmd, context: newContext(searchDirs = @[], partials = nbDoc.partials))
-  nbBlock.code = nbNormalize:
-    when defined(nimibPreviewCodeAsInSource):
-      getCodeAsInSource(nbDoc.source, cmd, body)
-    else:
-      toStr(body)
-  echo peekFirstLineOf(nbBlock.code)
+  if readCode:
+    nbBlock.code = nbNormalize:
+      when defined(nimibPreviewCodeAsInSource):
+        getCodeAsInSource(nbDoc.source, cmd, body)
+      else:
+        toStr(body)
+    echo peekFirstLineOf(nbBlock.code)
   blockImpl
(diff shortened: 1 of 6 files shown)


HugoGranstrom:
The implementation is done now. Opinions are welcome. I will add a few tests tomorrow :)

pietroppeter (changes requested):
only one change request, the rest looks good. thanks!

HugoGranstrom:
Hmm I don't think I can see the change request 🤔

src/nimib/blocks.nim
```diff
@@ -16,15 +16,16 @@ func nbNormalize*(text: string): string =
   text.replace("\c\l", "\n").replace("\c", "\n").strip # this could be made more efficient
 # note that: '\c' == '\r' and '\l' == '\n'
 
-template newNbBlock*(cmd: string, nbDoc, nbBlock, body, blockImpl: untyped) =
+template newNbBlock*(cmd: string, nbDoc, nbBlock, readCode, body, blockImpl: untyped) =
```
pietroppeter, line 19: `readCode` should be bool and we could position it after `cmd: string`. We probably could also add the same old api where readCode is defaulted to true (not sure if template composition runs into issues there...)

pietroppeter:
> Hmm I don't think I can see the change request 🤔

weird, cannot find it either. Added it as a single comment

src/nimib/blocks.nim
HugoGranstrom: Moving it to the front sounds good 👍 I'm a bit afraid of adding types to templates now after battling with them yesterday. (saw that you had also come across [this](https://github.com/nim-lang/Nim/issues/16680)) but as you already have a `string` there it should be fine. Do you mean to default it as a `readCode: bool = true` or making a separate overload without it that just calls this new one? That's also something I battled with yesterday as optional parameters aren't supported in templates with untyped arguments without the `experimental` pragma. So the separate overload is really the only option. (if template composition behaves as we want it to as you said 🤣)

HugoGranstrom:
> weird, cannot find it either. Added it as a single comment

Forgot to "add" it perhaps? Have happened to me before that I write a comment and then click the "Review Changes" button without actually "adding" the comment.

src/nimib/blocks.nim
pietroppeter: Yep separate overload. I also had issues with optional arguments in templates.

HugoGranstrom: I'll give it a try after lunch. Feels good to have a test suite in nimib to rely on with these things now :)

pietroppeter: Yes that is another Boone of recent PR... next step should be automatic generation of documentation with netlify/vercel preview as in scinim/getting-started

pietroppeter: Yes please

HugoGranstrom: Yes please to which one of them? 😅 Old or new?

Issue #82, opened 2022-03-21
[Idea] Support Python code blocks through NimPy


HugoGranstrom:
While procastinating more important things, a thought popped up in my head. What if one could use nimpy to get python code blocks in nimib? My main motivation for it would be creating presentations using nimiSlides for courses where Nim isn't an option. 

The idea is basically to abuse Python's `exec` to execute strings of python code, and it works across code blocks as well. Here's a very hacky implementation:
```nim
import nimpy
import nimib, nimib/capture

nbInit()
let py = pyBuiltinsModule()

nb.partials["nbPython"] = """
<pre><code>{{pyCode}}</code></pre>
<pre><samp>{{&output}}</samp></pre>
"""

template nbPython(pythonStr: string) =
  newNbBlock("nbPython", false, nb, nb.blk, pythonStr):
    nb.blk.context["pyCode"] = pythonStr
    captureStdout(nb.blk.output):
        discard py.exec(pythonStr)

nbPython: """
a = 6.28
print(a)
"""

nbPython: """
b = a + 1
print("b =", b)
"""

nbSave()
```

This is the result:
![bild](https://user-images.githubusercontent.com/5092565/159318875-a6ac87e7-7ebe-47a3-818e-d58a5dbec3cc.png)

It would also be trivial to add python highlighting to the string the same way we do it for markdown with `python"""python code here""""` in nimiBoost. 

Highlighting is lacking of course but #68 already exists. 

I'm not sure of the usefulness of this nor if there are any limitations that prevents it from being useful. I just thought it was a fun idea that actually seems to work.

pietroppeter:
Oh, I definitely like this and the implementation looks fine to me. I would a `nbInitPython` that hides all the initial implementation details (imports nimpy , create py, it could probably also add the template nbPython). Honestly, one additional thought I had was: can I have a nbPython block that takes as input Nim syntax? Is Python syntax a subset of Nim syntax? Maybe it does not work but in the cases it does it would be cool to get away without the string...

HugoGranstrom:
Nice! Will add an init template 👍

Regarding string-less Python:
- I would assume an untyped parameter could swallow basically everything so it is properly possible.
- it would probably need to use codeAsInSource as I don't trust astToStr to handle it well.
[...]

PR #81, opened 2022-03-12
Release 0.3

pietroppeter:
todo:

- [x] bump version 0.3
- [x] rebuild docs
- [x] added newNbCodeBlock
- [x] added newNbSlimBlock
- [x] address nbText inside template issue (will need to set readCode to false), see https://github.com/HugoGranstrom/nimib-reveal/pull/5#issuecomment-1072511222
  - [x] nbText now does not read code
  - [x] new nbTextWithCode that does read code
  - [x] fixed tests
- [x] fix nbFile
  - [x] remove the untyped version
  - [x] add example document files.nim
- [ ] checkout nimibook and fix stuff: https://github.com/pietroppeter/nimibook/pull/50
  - [x] add config utils
- [ ] checkout scinim/getting-started and fix stuff: https://github.com/SciNim/getting-started/pull/47
- [ ] add test for nbFile
- [ ] add test for nbImage
- [ ] add also nbAudio and nbVideo (with tests)
- [ ] add also nbShell?
- [ ] document new stuff (and also stuff from 0.2 minor versions)
- [ ] rebuild docs (including penguins)

later:
- [ ] config file found should report path relative to initDir

notes:
- the untyped version of nbFile was not working correctly when tested with the files context, so it is removed. It was probably never used.
- added a new newNbSlimBlock and newNbCodeBlock

PR #83, opened 2022-03-24
nbPython

HugoGranstrom:
implements #82 

I haven't tested this extensively yet but the non-string variant seems to work. I have one problem though, I can't have two overloads with the same name. Because to find out if it is a string it has to typecheck the code, the python code... So we need to use two separate names for them. Either one of them is `nbPython` and the other is `nbPythonCode` or `nbPythonString`. Or both have the more verbose names. Which of them do you prefer? 

TODO:

- [x] Get string version working
- [ ] Write tests

### DIFF
--- a/tests/tnimib.nim
+++ b/tests/tnimib.nim
@@ -84,11 +84,7 @@ suite "nbClearOutput":
     check nb.blk.output == ""
     check nb.blk.context["output"].vString == ""
 
-template whenModuleAvailable(module: untyped, body: untyped) =
-  when (compiles do: import module):
-    body
-
-whenModuleAvailable(nimpy):
+when moduleAvailable(nimpy):
   nbInitPython()
   suite "nbPython":
     test "nbPython string":
@@ -97,7 +93,7 @@ s = [2*i for i in range(3)]
 a = 3.14
 print(s)
 print(a)
-  """
+"""
       nbPython: pyString
       check nb.blk.code == pyString
       check nb.blk.output == "[0, 2, 4]\n3.14\n"
--- a/tests/tnimib.nim
+++ b/tests/tnimib.nim
@@ -6,18 +6,48 @@ nbInit # todo: add a test suite for nbInit
 suite "nbText":
   test "single line text string":
     nbText: "hi"
-    check nb.blk.code == "\"hi\""
     check nb.blk.output == "hi"
 
   when not defined(nimibPreviewCodeAsInSource):
     test "single line text string with strformat":
       let name = "you"
       nbText: fmt"hi {name}"
-      check nb.blk.code == "fmt\"hi {name}\""
       check nb.blk.output == "hi you"
 
     test "multi line text string":
       nbText: """hi
+how are you?
+"""
+      check nb.blk.output == """hi
+how are you?
+"""
+
+    test "multi line text string with strformat":
+      let
+        name = "you"
+        answer = "fine"
+      nbText: &"""hi {name}
+how are you? {answer}
+"""
+      check nb.blk.output == """hi you
+how are you? fine
+"""
+
+suite "nbTextWithCode":
+  test "single line text string":
+    nbTextWithCode: "hi"
+    check nb.blk.code == "\"hi\""
+    check nb.blk.output == "hi"
+
+  when not defined(nimibPreviewCodeAsInSource):
+    test "single line text string with strformat":
+      let name = "you"
+      nbTextWithCode: fmt"hi {name}"
+      check nb.blk.code == "fmt\"hi {name}\""
+      check nb.blk.output == "hi you"
+
+    test "multi line text string":
+      nbTextWithCode: """hi
 how are you?
 """
       check nb.blk.code == "\"\"\"hi\nhow are you?\n\"\"\""
@@ -29,7 +59,7 @@ how are you?
       let
         name = "you"
         answer = "fine"
-      nbText: &"""hi {name}
+      nbTextWithCode: &"""hi {name}
 how are you? {answer}
 """
       check nb.blk.code == "&\"\"\"hi {name}\nhow are you? {answer}\n\"\"\""
--- a/tests/tnimib.nim
+++ b/tests/tnimib.nim
@@ -84,17 +84,21 @@ suite "nbClearOutput":
     check nb.blk.output == ""
     check nb.blk.context["output"].vString == ""
 
+template whenModuleAvailable(module: untyped, body: untyped) =
+  when (compiles do: import module):
+    body
 
-nbInitPython()
-suite "nbPython":
-  test "nbPython string":
-    let pyString = hlPy"""
+whenModuleAvailable(nimpy):
+  nbInitPython()
+  suite "nbPython":
+    test "nbPython string":
+      let pyString = hlPy"""
 s = [2*i for i in range(3)]
 a = 3.14
 print(s)
 print(a)
-"""
-    nbPython: pyString
-    check nb.blk.code == pyString
-    check nb.blk.output == "[0, 2, 4]\n3.14\n"
+  """
+      nbPython: pyString
+      check nb.blk.code == pyString
+      check nb.blk.output == "[0, 2, 4]\n3.14\n"
 
--- a/tests/tnimib.nim
+++ b/tests/tnimib.nim
@@ -85,3 +85,16 @@ suite "nbClearOutput":
     check nb.blk.context["output"].vString == ""
 
 
+nbInitPython()
+suite "nbPython":
+  test "nbPython string":
+    let pyString = hlPy"""
+s = [2*i for i in range(3)]
+a = 3.14
+print(s)
+print(a)
+"""
+    nbPython: pyString
+    check nb.blk.code == pyString
+    check nb.blk.output == "[0, 2, 4]\n3.14\n"
+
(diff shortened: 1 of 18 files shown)


Vindaar:
Probably not possible for some reason I'm not aware of, but the python string case is for string literals, no?

Why not turn it into a macro and dispatch based on the node kind? If the argument is a string literal use that, otherwise treat is as code?

I.e. something like this in properly done:

```nim
import macros

proc doStrStuff(n: NimNode): NimNode = discard

proc doAstCodeStuff(n: NimNode): NimNode = discard

macro nbPython(code: untyped): untyped =
  echo code.treerepr
  if code.len == 1 and code[0].kind in {nnkStrLit, nnkTripleStrLit}:
    result = doStrStuff(code[0])
  else:
    result = doAstCodeStuff(code)

nbPython:
  a = 5
  print(a + "5")

nbPython: """
  b = 5
  print(b + "123")
"""
```

pietroppeter:
> we need to use two separate names for them

although @Vindaar 's solution seems elegant I would keep it simple and use `nbPython` for the string based variant (main api) and use `nbPythonim` for the code variant. Indeed it is a weird variant that takes nim code and interprets it as Python code (and we might expect not to support full python).

pietroppeter:
> Probably not possible for some reason I'm not aware of, but the python string case is for string literals, no?

actually thinking about this made me realize that having a template that takes a string as input (or in this case an untyped body expression that evaluates as string) make possible to generate more complex python code than what would be possible with the string literal approach.

HugoGranstrom:
I think I'll agree with Pietro on this one, allowing the input to be a string of any kind, literal or not, is more powerful. I'm not really sure why you would want to generate python code, but I'm sure that someone will find a cool way of using it. 

I'm not really a fan of the name `nbPythonim` though, the `-nim` ending doesn't make sense to me from a user perspective. They won't be thinking "Oh I'll write my Python code and then it will be interpreted as untyped Nim code". They will more think "Oh, I'll just write my Python code in this code block". So I'd prefer something like `nbPythonCode` or `nbPythonCodeBlock` which is more descriptive of what it does.

HugoGranstrom:
Are you okay with me adding tests for this as well? (And making Python a test dependency practically)

HugoGranstrom:
Here's a sneak peek of how the two variants looks when highlighted:
bild

pietroppeter:
while we are talking about highlighting, a though on the api for that. one of the reason that nimib uses often the `nb` prefix is that I did not want to "pollute" the namespace with common identifiers that someone would want to use. Something that slightly bothered me was that `nimib / boost` diverges a bit from that (`md` might have been something that one would want to use to convert a markdown string to html). Now we plan to introduce a new identifier in nimib / boost called `python`...
On the other hand I see how powerful it can be the highlighting and I am actually realizing that people might want it in general in nim for html stuff (also in nimib could be used)... so what if we change the api of nimib /boost to be `hlMd`, `hlFmd`, `hlPy`, `hlHtml`, `hlJs`, ...) or something similar? it might also help making it clearer that it serves the purpose of highlighting... it would be of course a breaking change. and probably we should open a separate issue for a discusion about this..

HugoGranstrom:
That's very true, thought about that as well. `python` is a really bad name to call it. The `hlLang(F)` syntax is fine by me 👍 I could implement it here and now if needed so no need for a separate issue unless we suspect someone else to have opinions on it.

pietroppeter:
> Are you okay with me adding tests for this as well? (And making Python a test dependency practically)
> And adding nimpy as a dependency for nimib?

ah, righht! nimpy as a dependency of nimib I would prefer to avoid. is there a way to hide this kind of functionality so that it is not imported if the library is not available (a static check on import error, maybe even a `when compiles(import nimpy)` would be enough)? I think in the future something like that would be useful also for other stuff (e.g. something that shows a html table starting from a datamancer dataframe would need datamancer...). If we manage to do that we could test it in the same way (and in the CI it is fine to add python and nimpy as dependency, we already have ggplotnim and others...).

the other option would be to package this kind of functionalities as separate libraries but I guess if it can be avoided it would be better for now.

HugoGranstrom:
Yep it's possible to give an error if it can't find `nimpy`: https://forum.nim-lang.org/t/3752
I'll add it 👍 


> the other option would be to package this kind of functionalities as separate libraries but I guess if it can be avoided it would be better for now.

I thought about doing that at first, but then it turned out to be so little code so it didn't feel worth making it a separate package.

HugoGranstrom:
We have a problem for the non-string version. List comprehensions don't work:
```nim
nbPythonCode:
	s = [i for i in range(3)]
```
Gives error:
```
Error: expected: ']', but got: 'keyword for'
```

So it seems we've managed to go beyond the capabilities of the Nim Compiler 🙃 Do we still want to keep this or are we satisfied with the highlighting of strings using `hlPy`?

tests/tnimib.nim
```diff
@@ -85,3 +85,16 @@ suite "nbClearOutput":
     check nb.blk.context["output"].vString == ""
 
 
+nbInitPython()
```
pietroppeter, line 88: let's put also this behind a `whenIsAvailable(nimpy)` switch (should we have a `whenIsAvailable` template/macro somewhere?). I want to make sure that to run tests you only need the required dependencies (then CI will double check)

HugoGranstrom: Good idea 👍 I can't seem to find any good place to put small internal templates like this so it's in the test file for now unless you have a better location

pietroppeter: let's put it directly in `src/nimib.nim` (and use it also to hide - instead of giving error - both nbInitPython and nbPython). I plan to use it also elsewhere

pietroppeter:
> Do we still want to keep this or are we satisfied with the highlighting of strings using `hlPy`?

well, it is true that since as demonstrated it breaks so easily, it could be removed and added to some example doc as a DIY novelty element, but since it is already there let's keep it and maybe we will move it out later; it should be anyway made clear that is not something that we plan to support. if you feel you want to remove it go ahead.

tests/tnimib.nim
HugoGranstrom: Done ✔️

HugoGranstrom:
> well, it is true that since as demonstrated it breaks so easily, it could be removed and added to some example doc as a DIY novelty element, but since it is already there let's keep it and maybe we will move it out later; it should be anyway made clear that is not something that we plan to support. if you feel you want to remove it go ahead.

I don't feel happy about making it available to users if we won't support it. But if you want to keep it for the future for a curiosity post we could keep it but not export it. How does that sound?

HugoGranstrom:
Btw, the CI seems to choke on the `nbPython` test. It doesn't seem to execute the code correctly or something like that as `nb.blk.output` is empty 🤔 
```
[Suite] nbPython
[nimib] 11 nbPython: 
    /home/runner/work/nimib/nimib/tests/tnimib.nim(99, 26): Check failed: nb.blk.output == "[0, 2, 4]\n3.14\n"
    nb.blk.output was 
  [FAILED] nbPython string
```
Could you test locally on your machine and see if it works? It works for me under WSL but I haven't got nimpy to work properly on my Windows install.

pietroppeter:
> I don't feel happy about making it available to users if we won't support it. But if you want to keep it for the future for a curiosity post we could keep it but not export it.

probably better to just remove it. We know where to find it if we need. I just realized that also the "untyped" version of `nbFile` was not working and will be removed (see latest commit in #81 ).

> Could you test locally on your machine and see if it works?

well if it is only the CI throwing problems we can skip the test and open an issue to fix the CI for that later.

I also just realized that this is a PR against my release-0.3 branch. I actually plan to merge that branch since I want to try and use it in my nimibook and getting-started PRs (changing dependency to nimib@#head) to see if everything works. Then when everything works I will open a new PR with a new branch to finalize the release (and later will merge in the PRs for nimibook and getting-started with the correct dependency). In the meantime the nbPython could be merged on main independently.

HugoGranstrom:
> well if it is only the CI throwing problems we can skip the test and open an issue to fix the CI for that later.

Yeah as long as it works on actual machines that's fine. I'll boot up my old laptop with linux and double-check that it works on linux there as well. 

> I also just realized that this is a PR against my release-0.3 branch. I actually plan to merge that branch since I want to try and use it in my nimibook and getting-started PRs (changing dependency to nimib@#head) to see if everything works. Then when everything works I will open a new PR with a new branch to finalize the release (and later will merge in the PRs for nimibook and getting-started with the correct dependency). In the meantime the nbPython could be merged on main independently.

The target branch is changed. I'll modify the code to make use of the `slim` code block and resolve the conflicts today or tomorrow.

HugoGranstrom:
Regarding the CI, look at the `nimpy` page in getting-started: https://scinim.github.io/getting-started/external_language_integration/nim_with_py.html

Notice how the code `discard py.print("Hello world from Python..")` doesn't produce an output. It's basically the same problem as we have here. I've tried it on my linux laptop now and it works if you compile it from the command line. But if I try to compile it with nimiBoost it doesn't show the output at all. So I suspect that it has something to do with the shell it is executed in and whether `nimpy` writes to the same stdout as nimib is capturing from or something along those lines. In other words, the shell that VSCode creates to run the build command is somehow different from the ordinary terminal shell. And if that is the case I have no idea how to solve that. It basically makes all automatic building systems using CI not able to use `nbPython` properly.

HugoGranstrom:
I'd say this is ready for merging now that the nbPython test is disabled and the CI seems happy enough.

pietroppeter:
Thanks for this! Yes, the CI issue is likely due to some weird shell thing that does not allow to capture output correctly.

PR #84, opened 2022-03-27
replace nimble deps task with nimble install

metagn:
To see if it works


pietroppeter:
thanks, definitely better, not sure why I had that contrived task :)

PR #88, opened 2022-06-16
Compiling nim to javascript

HugoGranstrom:
This is still bit of a mess but it's a somewhat working mess at least. Here's a fully working example:
```nim
import std/strutils
import nimib

nbInit()

nbText: "Counter POC"

template counterButton(id: string) =
  let labelId = "label-" & id
  let buttonId = "button-" & id
  nbRawOutput: """
<p id="$1">0</p>
<button id="$2">Click me</button>
""" % [labelId, buttonId]
  nbJsCode(labelId, buttonId):
    import std/dom
    echo "Hello world!"
    var label = getElementById(labelId.cstring)
    var button = getElementById(buttonId.cstring)
    var counter: int = 0
    button.addEventListener("click",
      proc (ev: Event) =
        counter += 1
        label.innerHtml = ($counter).cstring
    )


counterButton("1")
counterButton("2")

nbSave
```

Feedback is welcome 😄 Especially on where to put all of this code, it feels like it deserves it's own file to not clutter `nimib.nim`

## TODO
- [x] Test client-side Karax to not have to write HTML in strings (currently blocked by https://github.com/nim-lang/Nim/issues/19904 I think) Can be worked around by using `nbNewCode` API with string `"include karax / prelude"` and adding body with `script.addCode`. Manually importing all karax modules does work, it's only the include which isn't working.
- [x] Clean up in the code
- [ ] Test it
- [ ] Test it
- [ ] Test it
- [ ] Test it

## Pietroppeter's suggestions
- [x] Move as much as possible to `nimib/jsutils`
- [x] Rename to `nbCodeToJs`, `nbCodeToJsInit`, `addCodeToJs`, `addToDocAsJs`.
- [x] Use `NbBlock` instead of `NbCodeScript`
- [x] Add rendering code for this block
- [x] Add `nbCodeToJsShowSource` which should change a switch to show the Nim code.
- [x] Use a random string for `kxi_id` instead of global counter. (define a nimib-specific rng in `nbInit`)
- [x] Don't use `randomCounter` anymore, just check if key in cachetable and reuse it.
- [x] Write `counters.nim` showcasing reusable widgets
- [x] Write `caesar.nim` showcasing input handling


HugoGranstrom:
I've been able to get a Karax template to work by doing the `include karax / prelude` in a string first and adding the untyped code afterwards:
```nim
import std/random
template karaxButton() =
  echo "CReating a new button!"
  block:
    let root {.inject.} = "root" & $rand(int.high)
    nbRawOutput: """<div id="$1"></div>""" % [root]
    let script = nbNewCode: hlNim"""
include karax / prelude   
"""
    script.addCode(root):
      var counter = 0
      proc createDom(): VNode =
        result = buildHtml(tdiv):
          p:
            text "Hello world"
          label:
            text "Counter: " & $counter
          button:
            text "Click me (karax)"
            proc onClick() =
              counter += 1
      setRenderer(createDom, root=root.cstring)
    script.addToDocAsJs()
```
The problem I have though which I didn't have in the raw-dom example at the top, is that the `counter` variables are shared between multiple instances of the button. So clicking one button increases the count of all other buttons as well. In the previous example, different variables were generated. 

So it's partially working if you don't have any global variables that you update.

HugoGranstrom:
Upon further experimenting by adding a third button to the first example is also exhibiting this same behavior. The reason it didn't at first was that for some reason the first time we call the macro the code is treated as untyped but the second time as typed. And this resulted in two slightly different codes which had different names in the js. But adding a third button made the second and third button share the same names in the js. So we will have to do some manual name mangling ourselves to make this work. The complicating part is that we only have a string in the end and not the AST anymore.

pietroppeter:
Well, one thing is that we could make explicit the limitation that you could have a single js or Karan section in the page and you should have all the logic there.

Another option that comes to mind (but I know very little about this and not so sure if it would work) is to see if an iframe would work in sandboxing different parts of JavaScript (found this ref, have not had time to read it in detail: https://web.dev/sandboxed-iframes/)

HugoGranstrom:
> Well, one thing is that we could make explicit the limitation that you could have a single js or Karan section in the page and you should have all the logic there.

That is way too limiting in my eyes, plus for Karax to work we *have to* compile the different blocks in separate files for them to get their own Karax instances (it was just recently implemented support for this, see https://github.com/karaxnim/karax/pull/221). So putting them each into a block doesn't work (I've tried and only one of them rendered).

> Another option that comes to mind (but I know very little about this and not so sure if it would work) is to see if an iframe would work in sandboxing different parts of JavaScript (found this ref, have not had time to read it in detail: https://web.dev/sandboxed-iframes/)

Perhaps, but I have an idea for how to do this. It involves a CT table to keep track of the variable names for all variables in a specific script. And then when a new script is created the table is reset. This way we would keep the flexibility but isolate each script.

HugoGranstrom:
My solution actually seems to work 🤯 Now Karax should be fully working (if used with string imports and includes).

HugoGranstrom:
@pietroppeter Now we can start discussion namings and where to put this code. Do you have any preferences?

pietroppeter:
Wow! I will need to digest that a moment in order to get reasonable advice on naming and stuff. You know that merging this makes you an official maintainer now? ;) I will need you to get whatever rights you need for the repo (or we could move this and other repos to a nimib organization).

Anyway, let me get to basics. The feature we want to provide with this is a way to embed Nim derived js scripts in a page AND this would work also for Karax's framework, correct?

Can you add an example doc (or more than one if it works better) where we can see both in action (standard Nim to js and an example with Karax)? You should in this case merge current master (or rebase on top of it) so that you have the new structure where the example sources for docs are in docsrc.

HugoGranstrom:
> Wow! I will need to digest that a moment in order to get reasonable advice on naming and stuff.

My only idea so far is that the templates are still kept in `nimib.nim` but `nimToJsString` and all macros associated with it are moved to a separate file. 

> You know that merging this makes you an official maintainer now? ;) I will need you to get whatever rights you need for the repo (or we could move this and other repos to a nimib organization).

What?! :O What an honor! 😄 Or does this just mean I can't just dump my code here and expect you to maintain it for me? 🤣 
If you choose to create a nimib org, I would gladly transfer both nimiBoost and nimiSlide to it. But for you it also has downsides, making it into an org would make it less obvious that you are the master-mind behind nimib for example. 

> Anyway, let me get to basics. The feature we want to provide with this is a way to embed Nim derived js scripts in a page AND this would work also for Karax's framework, correct?

Exactly! Allowing us to finally add interactive elements to the output without inlining javascript (nor HTML). If you want to add a button that does something, this is exactly what you should use. 

> Can you add an example doc (or more than one if it works better) where we can see both in action (standard Nim to js and an example with Karax)? You should in this case merge current master (or rebase on top of it) so that you have the new structure where the example sources for docs are in docsrc.

Will do! Does it suffice if I add the counter buttons I have shown here or do you want to add something more spectacular as well? 😄 I don't have anything specific in mind but if you do, I could try implementing it.

HugoGranstrom:
> agree on that. the non essential (non public api) should be moved to a nimib/jsutils.nim from what we should import (and export) onlt what is necessary to have it working.
> on naming:

Those names and structure sounds good to me 👍 

> I would also make another change and use the standard NbBlock instead of a new type NbCodeScript. In code field there should be the original nim code and in output the code to JS. You would need to add the render function to html backend (the command would be nbCodeToJs. I might not seem some issues that might prevent this from happening and anyway would like to hear your opinion on this.

Yes, that sounds like a good idea. So `nbCodeToJsInit` returns a `NbBlock` but it isn't until we do `script.addToDocAsJs` that it is added to `nb.blocks`? And then the logic currently in `addToDocAsJs` is moved to the HTML backend which populates the `blk.code` field. Do I understand it correctly?

> Incidentally this should make it easier to show the orginal nim code in notebook (something that you want to do in the interactivity document for example).

I'm not sure what you mean here. Or I think you mean that we want this to work:
```nim
nbCode:
  nbCodeToJs:
    echo "Hello world!"
```
but I'm not sure how this would help. For example, if we do:
```nim
nbCode:
  nbText: "Hello world"
```
The output I get is neither a code block nor a text block but the `echo`s we are doing in `newNbBlock`:
```
[nimib] 5 nbText: -> Hello world

[nimib] 5 nbText: -> Hello world
```
So nestability of blocks doesn't work regardless. 

> I still have to review more but ideally I would try to avoid as possible adding other global objects, for example kxi_id. In principle we could have a field in nb.context that is called kxinames and it could be an array of strings. Once a nbCodeToJs block is created a new kxiname is created (and added to nb.blk.context) by making it distinct from previous ones (check if $len(kxinames) is already there if not increase the counter and try again). But this is an idea to make kxiname in principle customizable (would there be any advantage?) and we would keep track of all previous names. Not sure if this overcomplicates stuff, just an idea, there might be a better way, the gist is: let's make it something not a global variable but part of nb object and of current block object. We could add a field but maybe it is just more convenient to use the context.

I agree that using a global variable isn't good in this case. Building upon the `NbBlock` route, we only need a unique `kxi_id` for each block so I could use the context and just increase it every time the render proc is called. I don't really see a point in making it customizable though. It will only be used in a single page so there is no risk of collisions unless you combine multiple pages of course. A solution to that would be to generate a random string for `kxi_id` instead. And if we make it long enough we shouldn't run any risks of collisions and don't have to store previous ones. Then we don't have to store anything for this. What do you think about this?

> I am missing at the moment what is the role of randomCounter but it could be treated in the same way (and maybe this is more essential and we want instead to add field to the NbDoc and NbBlock object not sure).

`randomCounter` is a CT variable so we can't use our RT `nb` variable for it. The reason I added it was that multiple bodies would cause collisions in the macrotable. For example `nbCodeToJs` in a template would have the same body each time we call the template. But I now realize that we only have to add it to the table once and then check if it is in the table or not. So we should be able to get rid of `randomCounter` entirely. 

> well, yeah. Once we merge this I would definitely need your help in maintaining it (as it is indeed the case for nimibCodeAsSource) so let's make it official :)
> regarding not being recognized for originating nimib I do not care very much and I see the project as long term sustainable only if others join the effort.

💯 Sounds like a good plan and wise words👌 I'll happily help out where I can. 

> I have seen you already added an interactivity document that looks fine, maybe too many details in there but we can streamline it later. The counter examples are fine, it could be a counters.nim and it shows how to create resuable widgets (which is where the additional complexity of ids is needed).
>Another though I had would be a caesar.nim where you have two textboxes and two buttons (layout could be a very simple linear layout) and the buttons implement the encode and decode of [Caesar_cipher](https://en.wikipedia.org/wiki/Caesar_cipher).

Hehe yeah I wasn't sure how much to say about the details, especially since `nbCode` didn't play well with `nbRawOutput` in `addToDocAsJs`. Maybe adding a separate section at the with the detailed parts so those who are interested could read it and the others aren't bothered.
`counters.nim` sounds like a separation from the interactive part. And `caesar.nim` does sound like a fun example to add as well :o 😄 Will be interesting to see which of Karax and pure dom that is easier for that one.

pietroppeter:
> > I would also make another change and use the standard NbBlock instead of a new type NbCodeScript. In code field there should be the original nim code and in output the code to JS. You would need to add the render function to html backend (the command would be nbCodeToJs. I might not seem some issues that might prevent this from happening and anyway would like to hear your opinion on this.
> 
> Yes, that sounds like a good idea. So `nbCodeToJsInit` returns a `NbBlock` but it isn't until we do `script.addToDocAsJs` that it is added to `nb.blocks`? And then the logic currently in `addToDocAsJs` is moved to the HTML backend which populates the `blk.code` field. Do I understand it correctly?

yes, correct

> > Incidentally this should make it easier to show the orginal nim code in notebook (something that you want to do in the interactivity document for example).
> 
> I'm not sure what you mean here. Or I think you mean that we want this to work:
> 
> [...]
> 
> So nestability of blocks doesn't work regardless.

yes, nestability of blocks is a separate issue.

what I meant is that we could provide a `nbCodeToJsShowSource` functionality that if called after a `nbCodeToJs` (or after `script.addToDocAsJs`) would make the original nim code to appear in the document. The api would look like this:

```nim
nbCodeToJs:
  discard
nbCodeToJsShowSource
```

and the implementation could be something like this part (to be added in [`useHtmlBackend`](https://github.com/pietroppeter/nimib/blob/main/src/nimib/renders.nim#L13), note the usage of existing partial `nbCodeSource`):
```nim
  doc.partials["nbCodeToJs"] = """
{{>nbJsScriptNimSource}}
{{>nbJsScript}}"""
  doc.partials["nbJsScriptNimSource"] = "{{#js_show_nim_source}}{{>nbCodeSource}}{{/js_show_nim_source}}"
  doc.partials["nbJsScript"] = "<script>{{&output}}</script>"
```
_(note: you will need to highlight nim code)_

and this part:
```nim
template nbCodeToJsShowSource =
  nb.blk.context["js_show_nim_source"] = true
```

> > I still have to review more but ideally I would try to avoid as possible adding other global objects, for example kxi_id. In principle we could have a field in nb.context that is called kxinames and it could be an array of strings. Once a nbCodeToJs block is created a new kxiname is created (and added to nb.blk.context) by making it distinct from previous ones (check if $len(kxinames) is already there if not increase the counter and try again). But this is an idea to make kxiname in principle customizable (would there be any advantage?) and we would keep track of all previous names. Not sure if this overcomplicates stuff, just an idea, there might be a better way, the gist is: let's make it something not a global variable but part of nb object and of current block object. We could add a field but maybe it is just more convenient to use the context.
> 
> I agree that using a global variable isn't good in this case. Building upon the `NbBlock` route, we only need a unique `kxi_id` for each block so I could use the context and just increase it every time the render proc is called. I don't really see a point in making it customizable though. It will only be used in a single page so there is no risk of collisions unless you combine multiple pages of course. A solution to that would be to generate a random string for `kxi_id` instead. And if we make it long enough we shouldn't run any risks of collisions and don't have to store previous ones. Then we don't have to store anything for this. What do you think about this?

ah yes, probably better! in principle as an id you could even take `len(nb.blocks)` and in most cases this would work.

> > I am missing at the moment what is the role of randomCounter but it could be treated in the same way (and maybe this is more essential and we want instead to add field to the NbDoc and NbBlock object not sure).
> 
> `randomCounter` is a CT variable so we can't use our RT `nb` variable for it. The reason I added it was that multiple bodies would cause collisions in the macrotable. For example `nbCodeToJs` in a template would have the same body each time we call the template. But I now realize that we only have to add it to the table once and then check if it is in the table or not. So we should be able to get rid of `randomCounter` entirely.

even better :)

HugoGranstrom:
> what I meant is that we could provide a nbCodeToJsShowSource functionality that if called after a nbCodeToJs (or after script.addToDocAsJs) would make the original nim code to appear in the document. The api would look like this:

Ah ok, then I understand what you mean. Thanks for the tips :D

> ah yes, probably better! in principle as an id you could even take len(nb.blocks) and in most cases this would work.

If two scripts are created right after each other and then the `addToDocAsJs` for both are run after that, I think they will get the same id if we do it this way. As the number of blocks doesn't change until we actually add the scripts to `nb.blocks`. This is actually a quite likely scenario if two different templates are building up their scripts piece by piece using `addCodeToJs`. 

I'll start implementing all this tomorrow

HugoGranstrom:
I think that was it for the implementation part. The only thing I don't like is that the output of `nbCodeToJsShowSource()` shows the transformed code and not the original one. But even if I used the original body it would be a bit confusing to know what variables are captured or not as that part isn't included in the body (it's part of the `nbCodeToJs(cap1, cap2)`). But it's of course better than what we presently have

pietroppeter:
quick review, looks good. will review better later!

HugoGranstrom:
The `counters` example is mostly done but the code formatting is weird. I'd like to be able to reuse [this](https://github.com/pietroppeter/nimib/blob/6425743cf45f013df95914d6262749264d40cea7/src/nimib/blocks.nim#L23-L27) code to get the string of a body depending on if asInSource is set or not. I'll investigate whether that is possible tomorrow

pietroppeter:
Counters example is great and perfectly explained! Maybe the only thing I would add is one more copy of the button to show that the widget is indeed reusable and the counters do not interact with each other.

...and maybe you could add a reset button? I really feel the urge of resetting that counter (without refreshing the page) so bad! Although to be honest that counter is a very good minimal example to explain the feature! 😄

pietroppeter:
And the AsInSource would it helpful if we define `nimibPreviewCodeAsInSource` in the config for documentation? I think we should do that anyway.

HugoGranstrom:
> Counters example is great and perfectly explained! Maybe the only thing I would add is one more copy of the button to show that the widget is indeed reusable and the counters do not interact with each other.

Good idea! Will definitely add that 👍 

> ...and maybe you could add a reset button? I really feel the urge of resetting that counter (without refreshing the page) so bad! Although to be honest that counter is a very good minimal example to explain the feature! 😄

Haha, we could leave it as an exercise for the user to do that at the end 😜 And then they can look at the code in the `Show source` section. For Karax the change is minimal but for the pure dom version, I think it gets a bit too cluttered to add it to the tutorial. Is this a good compromise?

HugoGranstrom:
> And the AsInSource would it helpful if we define nimibPreviewCodeAsInSource in the config for documentation? I think we should do that anyway.

Adding it as a field to `nb`? Yes that sounds like a good idea 👌 

What I propose is basically being able to replace the linked code with:
```nim
nbBlock.code = getCode(nbDoc, cmd, body)
```
so that `getCode` can be reused. In my case I want to use it to get a prettier code output instead of always using `toStr(body)` as I'm doing now basically

HugoGranstrom:
> so great to see [caesar](https://e3a3c43c3015129fe6f8--nimib.netlify.app/caesar.html) come to life! 😍😍😍
output and source are so nice and clean!

Yes it turned out pretty neatly, the water.css styling is pretty convenient :D Are you happy with this or do you want explaining text as well? 🤣 

> maybe rename "key" to "shift", although I am not really sure. Key is indeed correct but in the context of caesar maybe shift is more easy to understand. thoughts?

Doesn't matter to me really, so we can go with shift if that is what's most commonly used in the Caesar context👍

>  range it from 0 to 25 with default 3 (the standard default).

Will do!

pietroppeter:
could not resist! https://twitter.com/pietroppeter/status/1542484628531019776?s=20&t=GZdX9XAmh1pGTka2hj8OVQ

pietroppeter:
> rb cqrb qxf byh wncfxatb jan oxavnm? cqnh yaxpajv j lryqna jwm ljw'c bcxy yujhrwp frcq rc? gm
bu, gurl qrsvagryl qb, ohg gura gurl punatr gur xrl, yrg'f fnl gb n cevzr ahzore, naq vg orpbzrf zber qvssvphyg!

> I will do some cleaning up of the code today but after that, I'm pretty happy.

I am **more than happy!**

just remember to add a line in index with a (relative) link and a sentence of explanation for all new examples (interactivity, counter, caesar). 

there was this old remark on counters, not sure if it was done or you did not see:
> Maybe the only thing I would add is one more copy of the button to show that the widget is indeed reusable and the counters do not interact with each other.
>
> ...and maybe you could add a reset button? I really feel the urge of resetting that counter (without refreshing the page) so bad! Although to be honest that counter is a very good minimal example to explain the feature!

other than, if there is someway to add simple test this it could be nice.

pietroppeter:
I was thinking, in the implementation we have a random generator to generate random strings, but in principle what we want is not really random stuff but unique stuff. The way the random string is built makes it very unlikely to find collision but it is still possible.

What if we replace that part with something simpler? like adding a `nb.id: int` (private) field to `NbDoc` and a new proc:
```nim
proc newId(doc: var NbDoc): int =
  result = doc.id
  inc doc.id
```
In this way everytime it is called (even multiple times inside a block) you get a new id (and then you can use $).
And in general other stuff could use this api for other purposes.

HugoGranstrom:
> just remember to add a line in index with a (relative) link and a sentence of explanation for all new examples (interactivity, counter, caesar).

Will do! 👍 

> there was this old remark on counters, not sure if it was done or you did not see:

I replied to with a suggestion somewhere above. Namely to keep the code less cluttered we leave it as an exercise to the reader to do, but they can look in `Show source` to see the Karax example. Look at the bottom of https://cfc027fee34d7ec235c5--nimib.netlify.app/counters

> other than, if there is someway to add simple test this it could be nice.

I'll investigate, it should be possible. Although the generated javascript is way too long to do a `==` on. 

> I was thinking, in the implementation we have a random generator to generate random strings, but in principle what we want is not really random stuff but unique stuff. The way the random string is built makes it very unlikely to find collision but it is still possible.
>What if we replace that part with something simpler? like adding a nb.id: int (private) field to NbDoc and a new proc:

That is a really clever idea! 🤯 This fits everything I need a random id for. We could replace all the ids in the examples with `nb.newId()` as well :D I'll remove `nb.rng` altogether then, if we need it in the future it is easy enough to add back in.

HugoGranstrom:
I forgot the cipher! 
> bu, gurl qrsvagryl qb, ohg gura gurl punatr gur xrl, yrg'f fnl gb n cevzr ahzore, naq vg orpbzrf zber qvssvphyg!

irel zhpu fb! gunax tbq gurer bayl ner 25 qvssrerag xrlf urer gubhtu!

HugoGranstrom:
I think this commit should fix the links in the preview as well by setting `docs` to `"."` if we are generating HTML and the full URL if we are generating markdown.

pietroppeter:
Yet another thought. Seeing how poweful the Karax use case is we could provide additional ergonomics.

Instead of:
```nim
let rootId = "karax-0"
nbRawOutput: "<div id=\"" & rootId  & "\"></div>"""
nbCodeToJs(rootId):
  import karax / [kbase, karax, karaxdsl, vdom, compact, jstrutils, kdom]

  karaxCode  # some code

  proc createDom(): VNode =
    result = buildHtml(tdiv):
      karaxHtmlCode # html karax code
  setRenderer(createDom, root=rootId.cstring)
```
we could have something like:
```nim
nbKaraxCode:
  karaxCode  # the karax code as above
  karaxHtml:
    karaxHtmlCode  # karax html code as above
```

`nbKaraxCode` could be a template that generates an id, the raw div and also provides inside itself the template `karaxHtml` (available only inside `nbKaraxCode`).

HugoGranstrom:
Ok, hit a problem immediately with `nbKaraxCode` 😅 Namely capturing variables. In your example the only variable we are capturing is `rootId` but the user might want to provide their own. So I'm afraid we will need yet another macro to separate out the captured variables from the body. But it should be doable, just not as elegantly.

HugoGranstrom:
The difference in boilerplate is so big! It just looks so clean 🤩 Have got it working (I think), but had to move the `karaxHtml` template to the javascript block in the macro. 

```nim
template karaxButton2() =
  nbKaraxCode:
    var counter = 0
    karaxHtml:
      label:
        text "Counter: " & $counter
      button:
        text "Click me (nbKaraxCode)"
        proc onClick() =
          counter += 1

template karaxButton() =
    let rootId = "karaxButton_root" & $nb.newId()
    nbRawOutput: """<div id="$1"></div>""" % [rootId]
    nbCodeToJs(rootId):
      import karax / [kbase, karax, karaxdsl, vdom, compact, jstrutils]
      var counter = 0
      proc createDom(): VNode =
        result = buildHtml(tdiv):
          label:
            text "Counter: " & $counter
          button:
            text "Click me (karax)"
            proc onClick() =
              counter += 1
      setRenderer(createDom, root=rootId.cstring)
```

pietroppeter:
So it seems all done, right? I am also repeating myself but it really is awesome!!!

My last (hopefully) proposal would be to add some finishing touches to Caesar. We could try to use `textarea` instead of `input` (should give a bigger area to write into) for plaintext and ciphertext, and add placeholder for both. For plaintext the placeholder could be something like "You can encrypt this message or you can try to decrypt the message below..." and for cipher text, well it would be some secret message you decide! :D

HugoGranstrom:
I think the only thing that's missing from your suggestions are some tests. But it's hard to do because the output field isn't populated until after we run nbSave as it's part of the render plan. So I guess it's possible to do but I don't know if it gives much more than using them in the docs and hope that the fact that the generated code actually compiles. 

Good ideas! 👌 I'll implement them tomorrow, shouldn't be much work.

HugoGranstrom:
> I think the only thing that's missing from your suggestions are some tests. But it's hard to do because the output field isn't populated until after we run nbSave as it's part of the render plan. So I guess it's possible to do but I don't know if it gives much more than using them in the docs and hope that the fact that the generated code actually compiles. 
> 
> Good ideas! 👌 I'll implement them tomorrow, shouldn't be much work. I like the "try to solve this" vibe btw 😆

HugoGranstrom:
I added some very rudimentary tests just for the sake of it to make sure the macros are running at least. 

Now that we have `nbKaraxCode` as an official template, should we add karax as a dependency I guess? In that case it will be `karax@#head` because they haven't released a new version in ages...

pietroppeter:
> I like the "try to solve this if you can" vibe btw 😆

Don't tell me that or I won't stop! (I am enjoying it too much! I guess you can tell this feature is something I have had in the back of my mind for too long...) 🤣 

Jokes aside, an interesting add on (but I would really consider this for a follow up PR) would be to a have a `karaxApp` theme (to be used like `nbInit(theme=karaxApp)`, based on html and js and css that can be found in [`karun`](https://github.com/karaxnim/karax/blob/master/karax/tools/karun.nim) module. Basically this would make nimib a better tool for using `karax` than `karun`... (I am assuming that `karun` does the theming part of `karax` from my recollection and a very quick check at the code, surprised not to find anymore mentions of it in the `karax` readme!) and remove a lot of the attrition that one feels when using karax...

HugoGranstrom:
> Mmh, I think we should treat it as we do with nbPython and nimpy, wrap it in a when moduleAvailable*.

Ah, right had almost forgot about `nbPython` 🙊 

> The fact that karax has not had a release in ages is definitely bad but not our issue (we can add our voices to ask for a new release). We could mention in the documentation that you will need karax#head to run the examples but hopefully that kind of remark would go out of date soon.
> *I looked back in how it is done in nbPython expecting that the advantage was to have better error messages (e.g. a user tries to use nbInitPython without having nimpy installed and nimib tells it to install the module) but indeed the user will not be able to access that. I will open an issue to improve that. edit: https://github.com/pietroppeter/nimib/issues/95

Agreed! Add a note about it for now and improving error messages sounds like a reasonable course of action 👍 

> Don't tell me that or I won't stop! (I am enjoying it too much! I guess you can tell this feature is something I have had in the back of my mind for too long...) 🤣

Haha 🤣 I can't stop you from doing things you like 😉 

> Jokes aside, an interesting add on (but I would really consider this for a follow up PR) would be to a have a karaxApp theme (to be used like nbInit(theme=karaxApp), based on html and js and css that can be found in [karun](https://github.com/karaxnim/karax/blob/master/karax/tools/karun.nim) module. Basically this would make nimib a better tool for using karax than karun... (I am assuming that karun does the theming part of karax from my recollection and a very quick check at the code, surprised not to find anymore mentions of it in the karax readme!) and remove a lot of the attrition that one feels when using karax...

That theme could work but I don't think it will beat writing karax in a dedicated file. In `nbKaraxCode` we do get autocomplete but we don't get error checks so you have to run the code to get the errors in the console. It is definitely easier if you just want to throw together a proof of concept of a small web app though so it would still have it's use-cases.

HugoGranstrom:
I've btw found a problem, namely importing local files. Because I move the `code.nim` to a temp folder when compiling it we have to add in the path of the original file which is easily fixed by doing a `--path:` when compiling. But it get's harder if the template is defined in a separate file from where it is called. For example if I define a template with a widget in nimiSlide and import it in a separate project. Then the separate project's path will be added but not nimiSlide's. And if the template is something like this:
```nim
template coolWidget*() =
  nbCodeToJs:
    import ./revealFFI
    # cool widget implementation
```
Then `./revealFFI` won't be found. So it would need to be able to register it's path. I have tried using `nb.context` but dealing with `seqs` there is a total mess as everything is converted to Values, so adding strings to it is convoluted to say the least. So I'm thinking of adding a `seq[string]` field to `NbDoc` to store them there instead. After this change the new code would be something like:

```nim
let (thisPath, _, _) = instantiationInfo(-1, true).filename.splitFile()
template coolWidget*() =
  nb.nimSearchPaths.add thisPath
  nbCodeToJs:
    import ./revealFFI
    # cool widget implementation
```

`nb.nimSearchPaths.add thisPath` would have to be added to every template which imports a local file. Do you have any better ideas? I don't think we can come around the fact that we must manually add the path though

HugoGranstrom:
The problem really is that we can't make it into a nice single template. Because if we get the path at RT the template will have expanded in the file that imported it instead. And if we get it at CT we can't define it in nimib because then it will just add the path of nimib instead. So the best option would probably be to give the user a template they can put in all files they need:
```nim
# This one is copied to each file needing it:
template addToSearchDirs() =
  let thisPath = currentSourcePath().parentDir()
  nb.nimSearchPaths.add thisPath

template coolWidget*() =
  addToSearchDirs()
  nbCodeToJs:
    import ./revealFFI
    # cool widget implementation
```

pietroppeter:
I have not followed all the arguments but what if instead of a temp folder we put code.nim (with a different name) next to the file we are processing? In this case if there is an error of compilation we do not remove it and debugging is easier.

HugoGranstrom:
I'm afraid that only solves one part of the problem. Not the part that is relevant for library makers. All libraries used wants to add their path to the search path, not only the file that is importing them

pietroppeter:
> Did that make it any clearer?

yes, I think this does make it clearer. It seems to me this puts a constraint on "widget developers" to use only modules which are available for a user of their library. It does not seem a completely unreasonable constraint to me.

In particular the nimislides "widget developer" (whoever that is 😝) should probably do:
```nim
# nimiSlide.nim (imported as package, so user.nim is not in the same folder)
template coolWidget() =
  nbCodeToJs:
    import nimislides / revealFFI
    foo()
```

where `revealFFI.nim` is in a `nimislides` folder next to `nimislides.nim` (same as someone using nimib can do `import nimib / jsutils`). `nimislides / revealFFI` is something that a user of `nimislides` can do having installed just nimislides.

does this put it in another context or do you think this is still a big issue for widget developers? (or maybe I am missing some other piece).

pietroppeter:
> That theme could work but I don't think it will beat writing karax in a dedicated file. In nbKaraxCode we do get autocomplete but we don't get error checks so you have to run the code to get the errors in the console.

You are right about that.

Getting back to this PR. We agree that theme we do it later?
If I think of stuff left to do, probably the only thing is to change the location (and name) of code.nim, right? I think it still make sense to put it next to the processed file (in case it accesses statically some assets) and not remove it if js compilation gives error.

For the name we could use:
- for `nbKarax`: `nbkarax3_myapp.nim` (where `myapp.nim` would be the original name of file and 3 one of the id that could be reused).
- for `nbCodeToJs` (and related): `nbcodetojs7_myapp.nim` (same as above)

HugoGranstrom:
> wheeew, I was sweating the imposter syndrome fearing I got it wrong! 😅

No you just are thinking in less convoluted ways than me often. I'm making things a bit took complicated to what they have to be sometimes :D

> Getting back to this PR. We agree that theme we do it later?

Yes, the theme can come later. But tbh I actually prefer the autostyling of water.css over manually having to assign classes using bulma (the CSS framework included by karun). 

> If I think of stuff left to do, probably the only thing is to change the location (and name) of code.nim, right? I think it still make sense to put it next to the processed file (in case it accesses statically some assets) and not remove it if js compilation gives error.

Yes, that sounds about right. And if we want to add `moduleAvailable(karax)` for now. Naming wise the renderer doesn't know the difference between `nbKaraxCode` and `nbCodeToJs`, they are both just `nbCodeToJs` to it. So the naming convention would simply be `nbcodetojs7_myapp.nim`. 

Btw, I got the use-case for which I wanted to create all of this working now. `coolWidget` was actually `typewriter` for nimiSlides to get a typewriter writing your text char-by-char. 🥳

HugoGranstrom:
TODO (I've found a few bugs):

- [ ] Guard using `moduleAvailable(karax)`
- [ ] Place generated file next to orignal
- [ ]  Gensym symbols that aren't gensym'd in a template (proc, iterator, converter), otherwise they will get the same names in the final code.

HugoGranstrom:
It feels kinda pointless to have tests that aren't run by the CI because it doesn't have karax installed. Would you mind if I added so `nimble docsdeps` was run for the tests as well?

pietroppeter:
Yes sure

HugoGranstrom:
If it passes the build now and the resulting pages seem to function as they should, I'd say it's finished now 🎊

src/nimib/renders.nim
```diff
@@ -10,6 +10,22 @@ proc mdOutputToHtml(doc: var NbDoc, blk: var NbBlock) =
...
+  let (dir, filename, ext) = doc.thisFile.splitFile()
+  let nimfile = dir / (filename & "_nbCodeToJs_" & $doc.newId() & ext).RelativeFile
+  let jsfile = tempdir / "out.js"
+  writeFile(nimfile, blk.context["transformedCode"].vString)
+  let kxiname = "nimib_kxi_" & $doc.newId()
+  let errorCode = execShellCmd(&"nim js -d:danger -d:kxiname=\"{kxiname}\" -o:{jsfile} {nimfile}")
+  if errorCode != 0:
+    raise newException(OSError, "The compilation of a javascript file failed! Did you remember to capture all needed variables?\n" & $nimfile)
+  removeFile(nimfile)
+  let jscode = readFile(jsfile)
```
pietroppeter, line 25: should we remove the jsfile after we read from it?

pietroppeter (changes requested):
minor points

src/nimib/renders.nim
HugoGranstrom: It is in a temp folder and will be overwritten soon enough so it's not neccecary but for completeness sake we can do it.

docsrc/caesar.nim
```diff
@@ -0,0 +1,61 @@
...
+  var shiftString = "3"
+  karaxHtml:
+    label:
+      text "Plaintext"
+    textarea(id = plaintextId, placeholder = "You can encrypt this message or you can try to decrypt the message below...")
+    hr()
+    label:
+      text "Ciphertext"
+    textarea(id = ciphertextId):
+      text "oek vekdt jxu iushuj auo! weet meha! dem oek sqd uqj q squiqh iqbbqt qi q fhypu, okcco!"
```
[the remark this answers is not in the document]
HugoGranstrom: bedw bylu dycyr! dycyr veh fhuiytudj! dycyr 4uluh!

pietroppeter:
checked everything and also gave a quick read on the docs preview to see if there were regression. everything seems fine.
Again great work on this, it has been fun!
You should have been invited to the repo, do you want to test your new powers and do the honors by merging this? :)

pietroppeter (approved):

HugoGranstrom:
> checked everything and also gave a quick read on the docs preview to see if there were regression. everything seems fine.
Again great work on this, it has been fun!
You should have been invited to the repo, do you want to test your new powers and do the honors by merging this? :)

Did the same and everything seems fine. Yes it has been fun and we have more fun to await now that we have the power of javascript at our fingertips.

What an honor! ❤️ Which method of merging do you prefer for PRs? Saw that it was 54 commits so perhaps squashing is the best option?

pietroppeter:
yes, squash and merge (for any PR in general)

HugoGranstrom:
Said and done! That felt nice 🥳🎉

PR #92, opened 2022-06-26
add netlify preview to CI

pietroppeter:
instructions taken from: https://github.com/pietroppeter/nimib/issues/71

requirements:
* netlify account with NETLIFY_AUTH_TOKEN, NETLIFY_SITE_ID

- [x] sign up to netlify with github
- [x] connect a repo to deploy: nimib
- [x] set publish directory to docs/
- [x] generate access token NETLIFY_AUTH_TOKEN using [this link](https://app.netlify.com/user/applications#personal-access-tokens), taken in [deploy action](https://github.com/jsmrcaga/action-netlify-deploy) used
- [x] get NETLIFY_SITE_ID from Site Settings in Netlify
- [x] add and adapt ymls to PR
  - [x] netlify_pr_task
  - [x] netlify_build_docs
  - [x] netlify_deploy_preview

PR #93, opened 2022-06-27
change of hello to test pr preview (do not merge)

pietroppeter:
follow up from #92 to see if preview works

### DIFF
--- a/docsrc/hello.nim
+++ b/docsrc/hello.nim
@@ -4,7 +4,9 @@ import nimib
 nbInit
 
 nbText: """
-## Secret talk with a computer
+## Secret talk with a CI
+Will you show me this in a preview or not?
+
 Let me show you how to talk with the computer like a [real hacker](https://mango.pdf.zone/)
 and incidentally you might learn the basics of [nimib](https://github.com/pietroppeter/nimib).
 ### A secret message


pietroppeter:
almost there but not quite yet (netlify_deploy_preview never started). cc @HugoGranstrom

HugoGranstrom:
If I go into your general Actions menu of Nimib I can see a failed netlify_deploy_preview. Did you trigger it manually or did it run on its own? 
It could be that it wasn't visible because it failed somehow

HugoGranstrom:
And comparing the logs with a run from getting-started it seems like for some reason secrets.NETLIFY_AUTH_TOKEN and secrets.NETLIFY_SITE_ID aren't properly defined, causing the failed connection to Netlify. Could you doublecheck that you have added those secrets to your repo?

HugoGranstrom:
Oh, found an interesting thing. You have to define the secrets as repository secrets and not environment secrets for them to be available automatically.

pietroppeter:
wow, you do know your way around these things! :)
I did define the secrets as environment secrets, I will try again adding them as repository secrets.

HugoGranstrom:
Almost 😅 One step closer to it working at least. Wonder why the link is invalid

HugoGranstrom:
The link in the Action logs seem to work but not the one linked by the preview button:
Correct:   https://39f6b89145686d6b99c4--nimib.netlify.app/
Incorrect: https://39f6b89145686d6b99c4--pietroppeter-nimib.netlify.app/

pietroppeter:
yes, indeed that was the issue, I will change that in this PR. Some other things I noticed:
- I think I was able to remove netlify deployments and indeed I do not see anymore that big notification that sent to something non existent
- there are a bunch of useless check added by netlify, probably removing some of the notifications in netlify settings will remove them
- links in the preview (e.g. from index to hello or to penguins) do not work since they are absolute, they should be made relative. minor annoyance, should be done later (I'll add it to release 0.3 backlog)

PR #96, opened 2022-07-02
make nimibPreviewCodeAsInSource the default and introduce flag nimibCodeFromAst for previous default, fix #94

### LINKED ISSUE #94: make nimibPreviewCodeAsInSource the default and add nimibCodeFromAst
The official release of 0.3 seems very close now. It make sense it will include the great work by @HugoGranstrom on finally bringing js backend based interactivity to nimib (one of the big milestone in the roadmap). I think it might be a good moment also to make default the `CodeAsInSource` (until now available through flag `nimibPreviewCodeAsInSource`, also excellent work by @HugoGranstrom btw).

We know it has edge cases where it will fail but it does not seem to be an issue for code as it is usually formatted (and the alternative `CodeFromAst` also has some issue - which did raise bugs directly in nim). The current default option would then be made available behind switch `nimibCodeFromAst`.

My idea would be to make it the default, testing if issues arises in nimibook and scinim/getting-started, and if no issues arise keep it for 0.3.

Thoughts?

pietroppeter:
fix #94.

With upcoming 0.3 release, it might be a good moment also to make default the `nimibPreviewCodeAsInSource` switch
and introduce a new switch `nimibCodeFromAst` for old default.

We know it has edge cases where it will fail but it does not seem to be an issue for code as it is usually formatted (and the alternative `CodeFromAst` also has some issue - which did raise bugs directly in nim).
We will test if issues arises in nimibook and scinim/getting-started before making the official release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,6 @@ a recently declassified top-secret cryptoanalytic weapon:"""
 nbCode:
   func decode(secret: openArray[int]): string =
     ## classified by NSA as <strong>TOP SECRET</strong>
-    # so secret that they do not want me to tell you and they will remove this message!
     for c in secret:
       result.add char(c)
 
@@ -90,12 +89,12 @@ nbText: """
 
 nbCode:
   let msg = decode secret
-  echo msg
+  echo msg  # what will it say?
 
 nbText:
   fmt"_Hey_, there must be a bug somewhere, the message (`{msg}`) is not even addressed to me!"
 
-nbSave # use nbShow to automatically open a browser tab with html output
+nbSave
 
 ```
 
--- a/docsrc/hello.nim
+++ b/docsrc/hello.nim
@@ -27,7 +27,6 @@ a recently declassified top-secret cryptoanalytic weapon:"""
 nbCode:
   func decode(secret: openArray[int]): string =
     ## classified by NSA as <strong>TOP SECRET</strong>
-    # so secret that they do not want me to tell you and they will remove this message!
     for c in secret:
       result.add char(c)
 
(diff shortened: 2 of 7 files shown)


pietroppeter:
rebased and forced push to include the changes of #88. if the checks are fine it will be ready to merge.
changes are rather straightforward, I also realized that one of the test that was protected from `CodeAsInSource` it could run fine with it.

pietroppeter:
so, the big outstanding issue that I see with `codeAsInSource` is that is not able to digest multiline string except if it starts on next line. This does not affect anymore `nbText` (now all tests pass for it since it does not read the code and so there is no check on code) but it affects `nbTextWithSource`. This is used only in `numerical.nim` (end of document when showing how the markdown table is created - now it is fixed), although in principle we should have used it also in `cheatsheet.nim` which instead uses a different mechanism (which I think it is equivalent to `codeFromAst`).

I would merge this anyway since it is not a big issue and there is a workaround. I will create an issue to track this problem that `nbTextWithSource` has with multilines. Thoughts, @HugoGranstrom ?

HugoGranstrom:
Yes I want to remember that problem when coding it and the rationale for needing it to start on a new line was that a *code block* can't start on the same line as the command unless it's a single-line code block. But now when dealing with the whole codeblock as a string that isn't a valid assumption anymore. I don't immediately see how to solve this in a smooth way but it's certainly possible. I'll investigate this and open a PR for it when I've solved it. So you can go ahead and merge this 👍

pietroppeter:
opened an issue to track the above: #97

PR #104, opened 2022-07-04
Fix codeAsInSource

### LINKED ISSUE #97: incorrect code for nbTextWithCode with multiline string on same line as command
Compiling and running this file:

```nim
import nimib
nbInit
nbTextWithCode: """problem
solution"""
echo "nb.blk.code: ", nb.blk.code
```

last line of output is (when using `CodeAsInSource` now default):

```
nb.blk.code: """problem
```

while it should be (as it is when using `-d:nimibCodeFromAst`)

```
nb.blk.code: """problem
solution"""
```

fix to be done in code in `src/nimib/sources.nim`

HugoGranstrom:
It took a day but I think I've found a somewhat non-chaotic implementation that seems to work. I added a few tests but I will add a few more and manually go through the docs to see if it looks OK. 

fixes #97 

## TODO
- [ ] Handle comments at end of code blocks
- [ ] Go through the docs and check for errors


HugoGranstrom:
Ugh, locally it runs correctly but this is so weird, the lines the error is pointing to doesn't make any sense either. Also I'm not really sure why it included all the commits from the last PR as well :/

PR #105, opened 2022-07-04
Fix codeAsInSource

### LINKED ISSUE #97: incorrect code for nbTextWithCode with multiline string on same line as command

HugoGranstrom:
Let's see if this works better now...

It took a day but I think I've found a somewhat non-chaotic implementation that seems to work. I added a few tests but I will add a few more and manually go through the docs to see if it looks OK. 

fixes #97 

## TODO
- [x] Handle comments at end of code blocks
- [x] Go through the docs and check for errors
  - [x] Penguin
  - [x] numerical
  - [x] cheatsheet
  - [x] mostaccio
  - [x] interactivity
  - [x] counter
  - [x] caesar
  - [x] hello
  - [x] files
  - [x] pythno
  - [x] nolan

### DIFF
--- a/docsrc/interactivity.nim
+++ b/docsrc/interactivity.nim
@@ -27,6 +27,8 @@ nbCode:
   ## Uncomment this line:
   ##script.addToDocAsJs()
 script.addToDocAsJs()
+
+nbText: "The complete code is:"
 nbCodeToJsShowSource()
 
 
--- a/docsrc/interactivity.nim
+++ b/docsrc/interactivity.nim
@@ -27,9 +27,7 @@ nbCode:
   ## Uncomment this line:
   ##script.addToDocAsJs()
 script.addToDocAsJs()
-
-nbText: "The complete code is:"
-nbCodeToJsShowSource()
+nbCodeToJsShowSource("This is the complete script:")
 
 
 nbText: hlMd"""
--- a/docsrc/numerical.nim
+++ b/docsrc/numerical.nim
@@ -115,8 +115,7 @@ nbText: "To compute percentage error of each method at $y(5)$ I will use:"
 nbCode:
   proc pe(yApprox: float): string = fmt"{(100.0*abs(yApprox - y5) / abs(y5)):.3f}%"
   echo pe y1hn[^1]  ## used like this
-nbTextWithCode: # note that we need to start the multi line on next line
-  fmt"""
+nbTextWithCode: fmt"""
 The following table is built as a Markdown table in a `nbText` block.
 
 **Table 1.** Percentage errors
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -152,8 +152,10 @@ when moduleAvailable(karax/kbase):
     nbRawOutput: "<div id=\"" & rootId & "\"></div>"
     nbKaraxCodeBackend(rootId, args)
 
-template nbCodeToJsShowSource* =
+template nbCodeToJsShowSource*(message: string = "") =
   nb.blk.context["js_show_nim_source"] = true
+  if message.len > 0:
+    nb.blk.context["js_show_nim_source_message"] = message
 
 
 template nbClearOutput*() =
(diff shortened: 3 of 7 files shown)


HugoGranstrom:
Ok, that seems to have fixed it but why did it work locally without any errors... 🤷

pietroppeter:
Nice! Can you try also to remove the `when defined(nimibCodeFromAst)` in `t nimib.nim`?

HugoGranstrom:
Something smells weird here, when the `defined(nimibCodeFromAst)` was included no errors. But when it was removed it gives an error at the very end (after the last test has been run with OK). I read somewhere that a plain `Error: execution of an external program failed:` means that the program returned a non-zero error code but I have idea how that could be dependent on this when-statement :/

HugoGranstrom:
To make it even stranger, running it with `nim r` gives this error but doing ``nim c` and then running the executable doesn't give any errors

HugoGranstrom:
Oh wait I'm just blind, there was a failed test 🤦

pietroppeter:
ah, right, it is to be expected! In nim multiline string remove the first newline so CodeFromAst does not see it, while CodeAsInSource sees it as it is written. They are both right! Probably the easieast way to fix it would be a short template/proc used only in tests and added to the multiline strings involved here:

```nim
template normalizeMultilineStringLiteral(s: string) =
  when not defined(CodeFromAst):
    if s[0] == "\n":
      return s[1 ..< s.len]
```

pietroppeter:
didn't see you already had the fix. fine as it is for me!

HugoGranstrom:
> ah, right, it is to be expected! In nim [multiline string literal remove the first newline](https://nim-lang.github.io/Nim/manual.html#lexical-analysis-triple-quoted-string-literals) so CodeFromAst does not see it, while CodeAsInSource sees it as it is written. They are both right! Probably the easieast way to fix it would be a short template/proc used only in tests and added to the multiline strings involved here:

TIL :D

> didn't see you already had the fix. fine as it is for me!

It is the only case where it matters thus far so making a template only feels warranted when we have a few test that needs it 👍

HugoGranstrom:
I've gone over the docs now and to my eyes, it looks like it's working but I would appreciate a second opinion on that 😅

pietroppeter (approved):

docsrc/interactivity.nim
```diff
@@ -27,6 +27,8 @@ nbCode:
   ## Uncomment this line:
   ##script.addToDocAsJs()
 script.addToDocAsJs()
+
+nbText: "The complete code is:"
```
pietroppeter, line 31: ah, no, sorry we cannot do it like this. `nbCodeToJsShowSource()` must come after the block. checking the document the code block is now disappeared (it is showing the code block of `nbText`). let's just remove this

pietroppeter (changes requested):

docsrc/interactivity.nim
HugoGranstrom: Oh right... 🤦 It was part of the nbCodeToJs block. So we should add it to the partial instead then I guess?

pietroppeter: you are right! we should find the block and then create a new code block with the code content. or in principle for this case we could just specify before the first code block that we will show also the nim code that will be compiled to javascript.

pietroppeter: > I don't really understand this part.

that part means that we do not change anything in `nbCodeToJsShowSource` and we just add some text before the code blocks saying that the after the first code block, in the second code block we will show the nim source. just a patch and we delay the fix.

 > My understanding is that we throw the "nbJsScriptNimSource" partial out the window completely, and create a new code block in nbCodeToJsShowSource which takes the code from the closest nbCodeToJs block in nb.blocks?

yes this would be my second suggestion. but now that you say it like that, indeed I like better your suggestion of having a message (even a default message) to show in nbCodeToJsShowSource.

pietroppeter (approved):

docsrc/interactivity.nim
pietroppeter: yep, go ahead!

PR #108, opened 2022-07-06
fix newline at beginning of code block

HugoGranstrom:
Sigh... This should be the last one from getting-started at least now. 

Starting a code block with a newline tricked it into thinking the indentation of the block was zero and thus all code below it in the file belonged to the the block.


pietroppeter:
Well getting-started seems to be a good gym :)

HugoGranstrom:
Haha indeed 🤣 This also has the advantage that my commit counts is through the roof 🤭

PR #113, opened 2022-07-11
rerun nimble readme

pietroppeter:
forgot to rerun `nimble readme` at the end of #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ This document is generated though nimib both as an index.html file and as a READ
 you should be reading one of the two, for the other:
 
 * [README.md](https://github.com/pietroppeter/nimib)
-* [index.html](.)
+* [index.html](https://pietroppeter.github.io/nimib)
 
 Nimib was presented in [NimConf2021](https://conf.nim-lang.org),
 see [video](https://www.youtube.com/watch?v=sWA58Wtk6L8)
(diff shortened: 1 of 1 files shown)


HugoGranstrom:
The example usage code-block just says `discard` now, even though it should be overridden by `hello` :see_no_evil:

pietroppeter:
what do you mean, I should actually check what I am merging before merging!? 😝

pietroppeter:
although to be honest this was a bug that was already there last commit...

HugoGranstrom:
> what do you mean, I should actually check what I am merging before merging!? 😝

🤣 No that would be too much 🤪
And yes it was from the small md refactor probably. 

If it wasn't for the fact that the readme soon will be independent from index.nim I would have proposed to add it to the preview somehow 🤭

pietroppeter:
> 🤦‍♂️ I'll go drag something old over myself 🤣

I gotta ask, never heard of this expression, is it possibly derived from Swedish? 🧐

HugoGranstrom:
I looked it up and yes it seems like it's not used in English 😅 after a bit of googling if found the English idiom "I'll sink through the floor" which is somewhat equivalent, you are embarrassed so you want to hide. In the Swedish case you hide under something old for some reason 😳

pietroppeter:
ah, nice! In italian it would be "sprofondare dalla vergogna" which would literally be translated as "sink in shame".

Issue #122, opened 2022-08-02
Error with `nbCodeToJs` and code that needs to be executed exclusively on js backend


pietroppeter:
example:

```nim
import nimib

nbInit
nbCodeToJs:
  import std/jsffi
```

trying to compile the above an error is produced:
```
~\Documents\GitHub\nblog\x_import_jsffi.nim(4, 1) template/generic instantiation of `nbCodeToJs` from here
~\Documents\GitHub\nimib\src\nimib.nim(145, 30) template/generic instantiation of `nbCodeToJsInit` from here
~\Documents\GitHub\nimib\src\nimib.nim(129, 43) template/generic instantiation of `nimToJsString` from here
~\Documents\GitHub\nimib\src\nimib\jsutils.nim(189, 26) template/generic instantiation of `checkIsValidCode` from here
~\.choosenim\toolchains\nim-1.6.6\lib\js\jsffi.nim(36, 10) Error: fatal error: Module jsFFI is designed to be used with the JavaScript backend.
```

indeed in jsffi there is a fatal error thrown if not running with js backend: https://github.com/nim-lang/Nim/blob/version-1-6/lib/js/jsffi.nim#L35

and this is the call that raises error in nimib: https://github.com/pietroppeter/nimib/blob/main/src/nimib/jsutils.nim#L189

thoughts on how to fix this @HugoGranstrom ?

pietroppeter:
the PR linked above is a workaround but probably there is a better solution.

HugoGranstrom:
Oh boy, that's a hard nut to crack :/ The entire reason we do the `checkIsValidCode` is to find out if it potentially could be a string instead of an untyped block of code. The best of course would be if we could capture this error but that feels unlikely.

An alternative workaround would be to auto-import all known JS-only modules similarly to how we insert `import std/json` [here](https://github.com/pietroppeter/nimib/blob/83cf3cf6e2cc1a46bdcc20c8f342f5422edead17/src/nimib/jsutils.nim#L108-L109).  But even then we would have problems with external js-only libraries not working.

HugoGranstrom:
Another workaround would be to inspect the code before the check, remove all imports, and then re-insert them again afterwards.

pietroppeter:
> The entire reason we do the checkIsValidCode is to find out if it potentially could be a string instead of an untyped block of code.

[...]

PR #123, opened 2022-08-02
add a define switch nimibJsutilsSkipCheckIsValidCode

pietroppeter:
workaround for #122


pietroppeter:
#122 is now fixed by #125

Issue #124, opened 2022-08-02
how to avoid mangling on names that need to be exported to js backend


pietroppeter:
as a simple example I was trying to test the following:

```nim
import nimib

nbInit
nbCodeToJs:
  import p5

  proc newColor(): Color =
    let
      r = random() * 255
      g = random() * 255
      b = random() * 255
      
    result = color(r, g, b)

  proc setup() {.exportc.} =
    createCanvas(500, 500)
    frameRate(3)
    
  proc draw() {.exportc.} =
    background(newColor())
nbSave
```

apart the fact that library [nimp5](https://github.com/Foldover/nim-p5) is currently broken (see https://github.com/Foldover/nim-p5/issues/1), the js that is created has mangled names:

```nim

import
  p5

proc newcolor_452984937(): Color =
  let
    r = random() * 255
    g = random() * 255
    b = random() * 255
  result = color(r, g, b)

proc setup_452984938() {.exportc.} =
  createCanvas(500, 500)
  frameRate(3)

proc draw_452984939() {.exportc.} =
  background(newcolor_452984937())
```

in order for this library to work, we would need to be able to avoid the mangling for names draw and setup (so that further js compilation tagged with export does not mangel them any further).

HugoGranstrom:
I feel like `nbCodeToJs` is falling apart like a house of cards :rofl: So the specific issue is that `exportc` isn't respected? Or is it mangled names in general?

pietroppeter:
related: https://forum.nim-lang.org/t/9341

HugoGranstrom:
> hey, we are using it! if it does not break it is not fun :stuck_out_tongue_closed_eyes: 

Haha true :rofl: If it wasn't for Karax we wouldn't have to mangle anything at all, but then we wouldn't have Karax :upside_down_face: 

> I would expect that exportc will be respected but the js we generate from original code has a mangled name and one of the purpose of exportc would be to avoid further mangling so that an external library is able to guess the correct name.

Okay, that I *should* be able to solve in the mangling logic :+1: 

> related: https://forum.nim-lang.org/t/9341

Worth noting is that as we can only have a single exported proc of each name, we are limited to 1 `p5` instance per nimib page.

pietroppeter:
> Worth noting is that as we can only have a single exported proc of each name, we are limited to 1 p5 instance per nimib page.

indeed in p5js there is a different mechanism if one wants to have multiple canvas on the same page:
[...]

PR #125, opened 2022-08-02
nbJs Maintainance

### LINKED ISSUE #122: Error with `nbCodeToJs` and code that needs to be executed exclusively on js backend

HugoGranstrom:
Fixes #122, ~~118~~, #124, #127

- [x] Separate untyped and string versions of `nbCodeToJs` (the body doesn't have to be type-checked anymore)
- [x] Rename nbCodeToJs to nbJsFromCode and nbJsFromString
- [x] Respect `{.exportc.}` when mangling names
- [ ] ~~Use separate mangling table for each script~~
- [x] nbKaraxCode postRender template

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,9 @@ When contributing a fix, feature or example please add a new line to briefly exp
 ## 0.3.x
 
 * fix "Did not parse stylesheet at 'https://unpkg.com/normalize.css/' in strict mode" (#120)
+* Split untyped and string versions of `nbCodeToJs` into `nbJsFromCode` and `nbJsFromString`. Same for `nbCodeToJsInit` → `nbJsFromCodeInit`, `nbJsFromStringInit` (#125)
+* Add `postRender` template to `nbKaraxCode` (#125)
+* `nbJsFromCode` now respects `exportc` pragma (#125)
 * _add next change here_
 
 ## 0.3 "Block Maker" (July 2022)
--- a/docsrc/counters.nim
+++ b/docsrc/counters.nim
@@ -6,7 +6,7 @@ nbInit
 nbText: hlMd"""
 # Counters - Creating reusable widgets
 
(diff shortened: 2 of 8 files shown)


HugoGranstrom:
It took a while but most of the things are done now. The problem of interlaced code blocks I will leave for a later PR as I'm not sure how to solve it best yet.

pietroppeter:
so basically this was ready some time ago but now it seems really ready :) should I give a quick review and merge?

HugoGranstrom:
Haha yes :) That would be appreciated, the most important thing being if I named everything new sensibly

pietroppeter:
as discussed in zoom call, we should add deprecation messages

HugoGranstrom:
I think this was what we agreed on. Correct if I'm wrong :)

PR #130, opened 2022-09-11
deprecate nbRawOutput for nbRawHtml and bump to 0.3.1

pietroppeter:
not sure if we discussed it @HugoGranstrom but I had in the back of mind also to deprecate `nbRawOutput` for a new `nbRawHtml` which does the same. `nbRawHtml` is the name that best describe the usage (while `nbRawOutput` is more about implementation detail of the block).

this PR also completes changelog for 0.3.1 and bumps nimble version, making it ready to release the 0.3.1.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,11 +9,15 @@ When contributing a fix, feature or example please add a new line to briefly exp
 
 ## 0.3.x
 
+* _add next change here_
+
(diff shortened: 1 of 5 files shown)


HugoGranstrom:
I'm all for this change, it has always felt a bit out of place 👍

PR #131, opened 2022-09-17
add hlHtml to nimiBoost

HugoGranstrom:
Adds HTML-highlighting support through nimiBoost

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ Nimib was presented in [NimConf2021](https://conf.nim-lang.org),
 see [video](https://www.youtube.com/watch?v=sWA58Wtk6L8)
 and [slides](https://github.com/pietroppeter/nimconf2021). 
 
-The vs code extension
-[nimiboost](https://marketplace.visualstudio.com/items?itemName=hugogranstrom.nimiboost)
-provides markdown highlighting in nimib file and a preview mechanism.
+The VS Codium / Code extension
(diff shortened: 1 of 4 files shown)


pietroppeter:
good for me, I just noticed that the mention of nimiboost in readme/index is:

```
The vs code extension [nimiboost](https://marketplace.visualstudio.com/items?itemName=hugogranstrom.nimiboost) provides markdown highlighting in nimib file and a preview mechanism.
```

Maybe it is worth mentioning formats other than markdown or reformulating that in another way (not sure how clear that sentence is for anyone that does not know what it means). Or maybe we update it later when we will add the link to Nimiboost livecoding video. ;)

HugoGranstrom:
Good catch, I've tried to improve it a bit. But it will be much easier after NimConf when we have a video to link to that shows what it can do. :)

pietroppeter:
Mmh, did you change in readme manually? We are still with index generated readme (will be removed), so changes should happen in index.nim and there should be a nimble task to update the readme. Other than that wording is much improved

HugoGranstrom:
Changed index.nim and ran `nimble readme`.

pietroppeter:
Ah ok, looking at the diffs I saw more changes in the readme than in index. Checking better I realize that last time we fixed nbCodeFromJs -> nbJsToCode the readme was not rerun and changes came out now. :)

Merging, thanks!

pietroppeter:
Ah ok, looking at the diffs I saw more changes in the readme than in index. Checking better I realize that last time we fixed nbCodeFromJs -> nbJsToCode the readme was not rerun and changes came out now. :)

Merging, thanks!

PR #132, opened 2022-09-17
Fix stdout windows issue

pietroppeter:
in the following example the terminal output is not captured on my windows machine (works on my mac machine). The fix in this PR (or uncommenting the flushfile line) makes it work for the windows machine:

```nim
import QRgen
import nimib

nbInit
nbText: "playing with  [QRgen](https://github.com/aruZeta/QRgen)"
nbCode:
  let myQR = newQr("https://github.com/pietroppeter/nimconf22-nimib")
  myQR.printTerminal
  #flushFile stdout  # needed on windows not on mac
nbSave
```

honestly I am not sure why this fix is needed, I just tried it because I had a hunch and it worked...

@HugoGranstrom: with latest commit I was able to make it work in captureStdout (very evident why... ;)), so no need to  document this anywhere, I guess.

I will add a line to changelog once I merge your changelog line from hlHtml PR (indeed our current changelog can run into conflicts, we could consider using something like https://github.com/iffy/changer)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@ When contributing a fix, feature or example please add a new line to briefly exp
 ## 0.3.x
 
 * Add `hlHtml` and `hlHtml` to nimiBoost
+* fix rarely occuring issue of terminal output not being captured (windows only, see example in https://github.com/pietroppeter/nimib/pull/132)
 * _add next change here_
 
 ## 0.3.1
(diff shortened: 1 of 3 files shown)


HugoGranstrom:
Ohhh yeah, that makes a lot of sense :rofl:

HugoGranstrom:
> I will add a line to changelog once I merge your changelog line from hlHtml PR (indeed our current changelog can run into conflicts, we could consider using something like https://github.com/iffy/changer)

Have read up on this now and this would be quite nice to have (only the fact that it bumps the .nimble version is nice :rofl:). At first I thought you had to create the markdown files manually, but seems like that is what the tool is doing for us.

PR #136, opened 2022-09-27
gensym captured variables that  haven't already been gensymed

HugoGranstrom:
If I tried to capture a global variable in `nbJsFromCode` inside a template, it didn't gensym the variable correctly (because it wasn't gensym'd by the template, the code didn't know to gensym it itself). This seems to fix it, though.

### DIFF
--- a/src/nimib/jsutils.nim
+++ b/src/nimib/jsutils.nim
@@ -136,9 +136,18 @@ proc genCapturedAssignment(capturedVariables, capturedTypes: seq[NimNode]): tupl
       import std/json
     for (cap, capType) in zip(capturedVariables, capturedTypes):
       let placeholder = gensym(ident="placeholder")
+      var newSym: NimNode
+      if "`gensym" notin cap.strVal:
+        newSym = gensym(NimSymKind.nskLet, ident=cap.strVal)
+        # add to tab[cap] = cap.gensym?
+        tabMapIdents[cap.strVal.nimIdentNormalize] = newSym.repr.ident
+      else:
+        newSym = cap
+      echo "In: ", cap.repr, " Out: ", newSym.repr
       result.placeholders.add placeholder
       result.code.add quote do:
-        let `cap` = parseJson(`placeholder`).to(`capType`)
+        let `newSym` = parseJson(`placeholder`).to(`capType`) # we must gensym `cap` as well!
+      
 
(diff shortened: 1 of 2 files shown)


pietroppeter:
Trust you on this one :), maybe add a change log line and then squash and merge ahead!

src/nimib/jsutils.nim
```diff
@@ -136,9 +136,18 @@ proc genCapturedAssignment(capturedVariables, capturedTypes: seq[NimNode]): tupl
...
     for (cap, capType) in zip(capturedVariables, capturedTypes):
       let placeholder = gensym(ident="placeholder")
+      var newSym: NimNode
+      if "`gensym" notin cap.strVal:
+        newSym = gensym(NimSymKind.nskLet, ident=cap.strVal)
+        # add to tab[cap] = cap.gensym?
+        tabMapIdents[cap.strVal.nimIdentNormalize] = newSym.repr.ident
+      else:
+        newSym = cap
+      echo "In: ", cap.repr, " Out: ", newSym.repr
```
Vindaar, line 146: Found a wild debug line :open_mouth:

HugoGranstrom: Hehe there are even more one like that, that got pushed in the release I realized :rofl: I'll clean them all up

HugoGranstrom:
> Trust you on this one :), maybe add a change log line and then squash and merge ahead!

Hehe, I barely understand it either, but it doesn't seem to brake any existing code and fixes broken code :upside_down_face:

pietroppeter:
do you happen to have a small example for the issue? it is probably not worth making a test about it but at least it have it here documented in the issue for when we have to refer back for some reason could be good.

pietroppeter (approved):

HugoGranstrom:
```nim
import nimib

nbInit

var counter: int # the gloabl variable in question
template component() =
  nbJsFromCode(counter):
    import std / dom
    window.addEventListener("load", proc (e: Event) =
      echo counter
    )
  counter += 1 # the counter should be different for the two components


component()
component()

nbSave
```
The `counter` should be different for the two components, but if you look in the JS console you see that both of them are printing `1` instead of `0` and `1`. This is because they are both named `counter_469762080` in the JavaScript so the second component's assignment, overrides the first one's. Hence why we have to gensym captured global variables.

Issue #118, opened 2022-07-14
Multiple `nbCodeToJsInit` can't be interlaced


HugoGranstrom:
As has been discussed between me and @pietroppeter, the file size of the final HTML file can grow quite a bit if many `nbCodeToJs` are used as they will have duplicate code between them that could be reduced. For this, we have the `nbCodeToJsInit` + `addCodeToJs` API. But there is one problem with it, namely that we are using a global table https://github.com/pietroppeter/nimib/blob/631ea5520917146ab92c6b5b11ee39fbd8257668/src/nimib/jsutils.nim#L12 which is reset each time `nbCodeToJsInit` is called (or `nbCodeToJs` which calls it in turn). Its job is to for every script, rewrite gensymed symbols to a form the compiler will accept in the external nim file. And to make sure different ones are rewritten to the same one, eg `a'gensym1` and `a'gensym2` are both rewritten to `a_1234567`. But if the table is reset it won't know that both of them should be rewritten to the same symbol. So this means this code won't work inside a template:

```nim
template foo() =
  let script = nbCodeToJsInit:
    let a = 1
  # a will in general be gensymed as it is defined in a template
  nbCodeToJs: # here the table is reset
    discard
  script.addCodeToJs:
    echo a # so it has no idea that a variable `a` was defined in the previous block here
```

The hard part is that this must be done at compiletime so we can't just assign a random id to each script, because the macro has no idea what script each code block comes from.

HugoGranstrom:
I haven't tested it yet but it could perhaps work if we pass in the script variable to `nimToJsString` and use it as a key for a table containing the existing table for each script. Each script variable, given that it isn't defined in a proc, should then get a unique key to the table. In theory... :sweat_smile:

pietroppeter:
to be honest this does not seem a big issue for me (but I understand optimistically half of what you say when you talk gensym, so maybe you are onto something good and keep going... ;)).

to recap, the discussion [was](https://github.com/HugoGranstrom/nimib-reveal/pull/5#issuecomment-1172887178) about the typewriter widget for nimislides, done with karax. My view on that is:
* karax should be used more for complete apps that take most of the page (and it is nice to be able to have multiple karax apps on a page mostly to document and explain karax).
* for widgets I would say it would be better to use vanilla ~js~ nim

[...]

PR #148, opened 2022-10-26
Refactor nbJs

HugoGranstrom:
As has been discussed in #118, the gensym-logic of nbJs is too complicated for the basic use-cases. Karax needs this advanced logic though, as each karax component must be compiled in its own file.

So the goal for this PR is to make the normal nbJsFromCode only use blocks instead gensym-logic. I have already started and I am getting nice behavior for simple cases. I need to check many more examples, though. 

This PR will introduce several breaking changes:
- `nbJsFromCode` will put the code inside a `block` so `import`s will not be allowed in them. Instead, you must use `nbJsFromCodeGlobal` which will be put at the top of the final script.
- `nbJsFromCodeInit`, `addCodeToJs` will be deprecated as we are limiting ourselves to isolated blocks. If you want interop between multiple `nbJs` blocks, define a global variable using `nbJsFromCodeGlobal`.

Now I'm going to try and fix the docs.

TODO:
from in-code review:
- [x] first 4 remarks are action to be taken, small details, when we close them we resolve the conversation
- [x] **decision to take**: do we interleave global scope and blocks or not?
  - Yes! We have 3 different kind of blocks including a interleaved global block and one placed at the top.
- [x] change logic of bump gensym, likely adding a counter and looping inside the macro. 👍 

rest of conversation.
- [x] **decision to take**: should we do something about possibility of showing code in JsFromCode?
  - Yes we will place each code block in its own NbBlock and use a collector block at the end. The old logic should still work for showing the code. 
- [x] move `compileNimToJs` from `renders.nim` to `jsutils.nim`
- [ ] Update Docs
  - [x] first example will not be with capture variables
  - [x] will add a `internal working` sections on sources of complexity
  - [ ] example on how to show appreciation about single js script with global and block sections.


HugoGranstrom:
I have tested this new slim nbJs on our documentation and our NimConf slides now and it seems to work. Tomorrow I'll go through your nblog posts which uses nbJs. Specifically there was one which needed `exportc` to work. Will be interesting to see how it works out

HugoGranstrom:
I checked nblog now and everything seems to work if we just move the imports to a global block :partying_face:

HugoGranstrom:
Documentation and tests have been updated now. It's actually quite impressive how short `jsutils.nim` is now that we have thrown away all the gensym logic :exploding_head: 

I actually think this PR might be ready for review :o @pietroppeter

pietroppeter:
Have you tried running the three_js example in nblog?

pietroppeter:
I am thinking that we might want to have also a block for when you plan to have a single (nim derived) script in the entire page. 
We could have a template that does not require a separate global block. Should be easy to add, right? name?  nbJsFromCodeSingle? nbJsFromCodeOnlyOne?

HugoGranstrom:
I don't quite understand what you mean. We have nbJsFromCodeOwnFile which karaxCodeBlock uses internally which compiles a code block the old way into its own file. Is that what you mean?

pietroppeter:
Without Gensym so that a case like threes_js works without a separate global section

HugoGranstrom:
Nothing uses gensym anymore. So for this case you could either use `nbJsFromCodeOwnFile` or `nbJsFromCodeGlobal` and just put all your code in there directly. Both ways you won't put the code inside a `block`.

pietroppeter:
ah ok, not even karax block use gensym? so two identical karax blocks will not run into issues of sharing variables?
anyway yeah I did not realize that I could just use `nbJsFromCodeGlobal` for everything.

and I need still to actually review code and understand for real this time how all this js stuff works. 🙄 
in the back of my mind I still have the old question: why can't I just write code in a separate nim file, compile with js backend and put the output in a script tag in the document? I know it is wrong for many reason, but I still not really understanding where the different pieces of current complexity come from (I know about colliding identifiers when using more than one script,  issue that become even worst when the script is basically the same - e.g. a widget, karax based or not, I know of captured variables requiring effort, not sure if there is other stuff).

thanks for all the gentle explanations! ❤️

HugoGranstrom:
> ah ok, not even karax block use gensym?

No, the `bumpGensym` I talked about yesterday in #118 is implemented and solved the problem :D It is gone forever! Look at the diff for `jsutils.nim` and enjoy all the red B)

> and I need still to actually review code and understand for real this time how all this js stuff works. roll_eyes

The code has never been as comprehensible as it is now, so now is a good time to dig in. Keep shooting questions at me, I'm not merging this until you understand it ;) 

> in the back of my mind I still have the old question: why can't I just write code in a separate nim file, compile with js backend and put the output in a script tag in the document? I know it is wrong for many reason, but I still not really understanding where the different pieces of current complexity come from (I know about colliding identifiers when using more than one script, issue that become even worst when the script is basically the same - e.g. a widget, karax based or not, I know of captured variables requiring effort, not sure if there is other stuff).

Most of the complexity is gone now thanks to the removal of all things gensym. So this is basically what we are doing now. Except that we are baking all the scripts into a single file (for ordinary `nbJsFromCode`) separated by blocks. This should improve build-times compared to compiling each `nbJsFromCode` as its own file (remember how slow `cornerImage` was in the `nimConfSlide`? It was exactly for this reason).
And the old complexity came basically solely from the colliding identifiers problem. So now that we have solved it (bumpGensym + blocks), it's not needed anymore. In theory, we could just use `bumpGensym` and keep the old behavior that all code blocks are totally isolated from each other and compile them one by one as you suggest. But the `block`s are simple enough and allow us to have interop between code blocks, so I think it is the superior way forward. 

Capturing variables is probably the most complex part that we have left, but it is "only" ordinary typed macro stuff to generate the correct code to serialize and deserialize the variable on the different sides.

src/nimib/renders.nim
```diff
@@ -16,7 +16,19 @@ proc compileNimToJs(doc: var NbDoc, blk: var NbBlock) =
...
+  var codeText = blk.context["transformedCode"].vString
+  if blk.context["isOwnFile"].vBool:
+    var bumpGensymString = """
+import macros
+
+macro bumpGensym() =
+  let _ = gensym()
+
+""" 
+    for i in 0 .. doc.newId():
```
pietroppeter, line 28: so this is maybe a bit odd, from what I understand.

you want a way to have an increasing number of bumpGenSym for each ownFile of js that you compile. So, our JsGlobalScript has no bumpGenSym, correct? it would be ok to have 1 bumpGenSym for first ownFile, 2 bumpGenSym for second ownFile, and so on?

if the above is correct, I think that here we are adding a "sort of random" number of gensym (depending on how many docId we have consumed up to that moment) for each ownFile. So the first ownFile might get 7 bumps, the second ownFile will get at least 8 but might get 11, and so on...

What if instead of this we add a jsOwnFilePrelude (or jsOwnFilePrefix) to NbDoc type that during init is initialized only with the macro bumpGenSym, then every time you compile a ownFile you add a single bump. In this way first file would get a single bump, second ownFile would get two bumps,  and so on... in principle it should not change much* but it feels a bit better.

*there is the off chance that having more bugs will prevent collision but if that is the case then we should every time add a big number of bumps (in principle we could add as many lines in code to be conservative...)

pietroppeter (commented):
added some comments (mostly questions) from what I understand on the changes. Given that I did not understand the code you are replacing I will also add more questions in general reviewing the complete files in your actual branch

pietroppeter:
Ok, I have read the rest and I can go on with comments and questions, it seems I understand now most of the code except the stuff in nimToJsString. Correct me on the statements below if I am wrong.

In `nimib.nim` we define basically 4 api to add js code generated from nim code:
  - `jsFromString`: consists of 4 templates (`nbJsFromStringInit`, `nbAddStringToJs`, `nbAddToDocAsJs` and `nbJsFromString`). This is the easiest to understand as it does what one would guess: put the content in a nim file, compile it to javascript and then add it in the document in the place where the block was placed. The only piece of complexity here is the bumpGenSym that we are doing and we will do for every nim own file that compiles to javascript. This is though an api that I think is never used in the documentation (correct?) as you would need to put nim code in a string which is kind of ugly. There is no capture of variables (you can do it manually by interpolating strings). Not sure where this api should be used in place of the others but it is a sort of escape hatch in case the other have bugs for some reasons...
  - `jsFromCodeOwnFile`: this is a first extension of the one before, you are now able to write nim code without putting it into a string AND you are also able to capture variables. Being an OwnFile you do the bumpGensym to avoid collisions of identifiers.
  - `nbKaraxCode`: this the Karax api, build on `jsFromCodeOwnFile` without the boilerplate (and adding the Karax id for multiple Karax blocks in the same file). It also captures variables.
  - `nbJsFromCode`: this is the one that uses the new intuition of a global js file. It assumes we have a single global script for every document. You can add stuff in the global scope or you can add stuff in a block scope. Great advantage is that all code in here gets compiled in a single script (so less time to compile and less duplicate js at the end). We are capturing variables of course. We do not actually add a block until there very end, currently this is put inside nbSave. Some thoughts about changing stuff here (besides the comments already done in the in code review):
    + what if instead of of `nbJsFromCodeGlobal` and `nbJsFromCode` we call them respectively `nbAddJsFromCode` and `nbAddJsFromCodeInBlock`? we are indeed adding js to a global script that has been initialized (as empty) during nbInit.
    + it is a bit weird that the block is added during nbSave. already in the in-code-review I propose to extract a proc. I would propose that we call it `addJsGlobalCodeBlock` or something similar. it could be called by the user as `nb.addJsGlobalCodeBlock`. Indeed we probably might want to add another field to NbDoc: `isJsGlobalCodeBlockAdded: bool`. So that in `nbSave` we check it and we add it only if it is false.
    + in this way we can add more html after this js code block (not sure if it would matter or not). But in particular something that we can do is to show the original code for js global code block with `nbCodeToJsShowSource` (should this template and the name of the command should be renamed from nbCodeToJs to nbJsFromCode? or maybe it is better like this since this block is actually shared also with the first api `fromString`?). I think currently we are losing the originalCode in `nbJsFromCode` blocks, we probably could add a new field to `NbDoc` to track "original" code.

On capturing variables:
- it seems that `nimToJsString` is the only big piece of complexity in `jsutils` (probably we could move `compileNimToJs` from `renders` into `jsutils`?), since `nbKaraxCodeBackend` is rather transparent. You confirm that this complexity is only for the captured variables? if this is the case could we in principle check with a `when len(args) == 1` (directly in `nbFromCode` blocks) and NOT call `nimToJsString`?
- regarding capturing variables and `nimToJsString` could you explain better what that code does? probably it would help if you show the original vs transformed nim code to see how it would look. The code is really complex with two stages, a cache, a call to degensym...

Finally I think, having understood a bit more, we probably could improve the interactivity document:
- the first example we show should not probably be one with a captured variable
- we should mention the multiple apis and what are the sources of complexity, which should be essentially two: nim gensyms identifiers when compiling to js, we want to capture variables. 
- regarding the multiple api the global js script api it seems to me is a novel and smart idea on how to solve the sources of complexity, so it should be given a bit more appreciation.
- also on capturing variable it is now evident to me (especially looking at the examples of two counters) that it is important for passing ids and have reusable widgets that contain both rawHtml and Js code. At first I thought its main use case was to pass complex objects generated from C, but I guess that is an additional (cool) use case (on which I do not think I have really seen examples, I mean in principle we could train a neural network with arraymancer and translating to a easier structure where we could do the inference on js... you know just to start with simple ideas :D).

pietroppeter:
Ok, with the in-code-review and last comment, I think I have exhausted my review on current code. Hey I am taking this literally 🤣:

> Keep shooting questions at me, I'm not merging this until you understand it ;)

src/nimib.nim
```diff
@@ -192,7 +206,21 @@ template nbSave* =
   # order if searchDirs/searchTable is relevant: directories have higher priority. rationale:
   #   - in memory partial contains default mustache assets
   #   - to override/customize (for a bunch of documents) the best way is to modify a version on file
-  #   - in case you need to manage additional exceptions for a specific document add a new set of partials before calling nbSave
+  #   - in case you need to manage additional exceptions for a specific document add a new set of partials before calling
+  if nb.nbJsGlobalScript.len > 0 or nb.nbJsScript.len > 0:
+    nb.nbJsGlobalScript = "import std / json\n" & nb.nbJsGlobalScript
+    let completeJsCode = nb.nbJsGlobalScript & "\n" & nb.nbJsScript
+    echo "Complete Js Code: \n", completeJsCode
```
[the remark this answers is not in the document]
HugoGranstrom: The `echo` at the bottom, yes.

src/nimib.nim
```diff
@@ -192,7 +206,21 @@ template nbSave* =
   # order if searchDirs/searchTable is relevant: directories have higher priority. rationale:
   #   - in memory partial contains default mustache assets
   #   - to override/customize (for a bunch of documents) the best way is to modify a version on file
-  #   - in case you need to manage additional exceptions for a specific document add a new set of partials before calling nbSave
```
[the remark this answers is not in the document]
HugoGranstrom: Oops, will fix that :+1:

src/nimib.nim
```diff
@@ -192,7 +206,21 @@ template nbSave* =
...
+  #   - in case you need to manage additional exceptions for a specific document add a new set of partials before calling
+  if nb.nbJsGlobalScript.len > 0 or nb.nbJsScript.len > 0:
```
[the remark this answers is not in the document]
HugoGranstrom: Absolutely, will do :+1:

src/nimib.nim
```diff
@@ -132,10 +132,12 @@ template nbRawHtml*(content: string) =
 template nbJsFromStringInit*(body: string): NbBlock =
   var result = NbBlock(command: "nbCodeToJs", code: body, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
   result.context["transformedCode"] = body
+  result.context["isOwnFile"] = true
   result
 
+#[
```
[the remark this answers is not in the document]
HugoGranstrom: Yes indeed :+1:

src/nimib/types.nim
```diff
@@ -33,6 +33,7 @@ type
     renderPlans*: Table[string, seq[string]]
     renderProcs*: Table[string, NbRenderProc]
     id: int
+    nbJsGlobalScript*, nbJsScript*: string
```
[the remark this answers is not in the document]
HugoGranstrom: We could do it with one script and interlace the block and global scripts like this:
```nim
global1
block:
  block1
block:
  block2
global2
global3
block:
  block3
```
But it would be a different behavior than putting all globals at the top:
```nim
global1
global2
global3
block:
  1
block:
  2
block:
  3
```
as the order of execution changes. So it depends on what we want the global code to be able to do. I like the global-at-the-top approach because it allows us to do customizations of global variables defined by other blocks. Let's say that we have a block which exposes a `config: Table`. Then to configure this block, you modify it inside a `nbJsFromCodeGlobal` block like this:
```nim
nbJsFromCodeGlobal:
  config["param1"] = 3.14
```
The place in the code where this variable change takes place is crucial for if it will take effect or not. If `block1` in the first example is where the hypothetical block reads `config` and starts whatever process it does. And if `global2` is where we change this parameter, then it will be changed *after* it was read and might not make a change. If on the other hand we had all the global at the top like in the second example, `config` would be changed in time before `block1` is run.

This is a very hypothetical example, but it shows a potential use-case which wouldn't be possible if we didn't separate the global and block codes.

src/nimib/renders.nim
HugoGranstrom: > you want a way to have an increasing number of bumpGenSym for each ownFile of js that you compile. So, our JsGlobalScript has no bumpGenSym, correct? it would be ok to have 1 bumpGenSym for first ownFile, 2 bumpGenSym for second ownFile, and so on?

Yes, yes and yes.

> if the above is correct, I think that here we are adding a "sort of random" number of gensym (depending on how many docId we have consumed up to that moment) for each ownFile. So the first ownFile might get 7 bumps, the second ownFile will get at least 8 but might get 11, and so on...

Correct. It makes it a bit of a pain to track down bugs, true. Adding a block which doesn't use nbJs but `newId()` could introduce an unfortunate shift indeed. 

> What if instead of this we add a jsOwnFilePrelude (or jsOwnFilePrefix) to NbDoc type that during init is initialized only with the macro bumpGenSym, then every time you compile a ownFile you add a single bump. In this way first file would get a single bump, second ownFile would get two bumps, and so on... in principle it should not change much* but it feels a bit better.

We could absolutely do that, or just have a counter for this. It hit me that we could do `bumpGensym(i)` and do a loop of `i` gensyms inside a single macro call instead. Having 100 calls to a macro (even though it is simple), might have some bad effects of build times, perhaps. So I'd go with a counter for this and moving the loop inside the macro instead.

HugoGranstrom:
>  (probably we could move compileNimToJs from renders into jsutils?

Sure, will do :+1: 

> You confirm that this complexity is only for the captured variables? if this is the case could we in principle check with a when len(args) == 1 (directly in nbFromCode blocks) and NOT call nimToJsString?

The complexity is mainly due to the capturing of variables, but there still is logic in there that we need all the time. We still need the `degensym` to clean ```varName`gensym32``` and the logic to put the code inside of a `block`. So I don't see any way of avoiding calling `nimToJsString`. If we really wanted to, we could perhaps make it so it didn't have to call the second stage macro in these cases, but I don't see why we would do that. It would only make the code more fragmented and harder to follow. 

> regarding capturing variables and nimToJsString could you explain better what that code does? probably it would help if you show the original vs transformed nim code to see how it would look. The code is really complex with two stages, a cache, a call to degensym...

The fact that we have two stages is that we want the body to be untyped. This is because typed code has been transformed and isn't the code that we wrote originally. Here's an example:
```nim
# untyped
let a = 1
let b = 2
echo a + b

# typed
let a = 1
let b = 2
echo [a + b]
```
As you can see, if we compiled the typed code in a file, it would echo out an array instead of an int. So having the body be untyped is very important. But we also want to get the type of the variables we want to capture, so that we can unserialize them to the correct type on the javascript side. So we need a typed macro to get those types. The question that is, how do we pass the untyped body to a typed macro without making it typed? The answer is putting it in a cache table! In the typed macro (second stage), we can then get the untyped AST of the body from the cache table. The key to the element in the table, `key: static string`, is passed as an argument to the typed macro. 

So this should hopefully bring some light on why we need the two stages, the cache table and degensym.

HugoGranstrom:
Now for the actual capturing of variables! 
```nim
let captureVars = toSeq(captureVars)
```
Here we save the symbols to the captured variables in a seq. Next we get the type of each symbol:
```nim
let captureTypes = collect:
  for cap in captureVars:
    cap.getTypeInst
```
Here we fetch the untyped body from the cache:
```nim
var body: NimNode
if key in bodyCache:
  body = bodyCache[key]
```
---

Next up we generate the assignments and deserialization of the captured variables:
```nim
# 1. Generate the captured variable assignments and return placeholders
let (capAssignments, placeholders) = genCapturedAssignment(captureVars, captureTypes)
```
If we for example have this code:
```nim
var x: int = 123
var y: string = "hello world"
nbJsFromCode(x, y):
  echo x, y
```
What `genCapturedAssignment` will give us is the following code:
```nim
let x = parseJson(placeholder`gensym1).to(int)
let y = parseJson(placeholder`gensym2).to(string)
```
and the list of placeholder-symbols it has generated. Now what's up with ``placeholder`gensym`` you might ask? We don't know the value of the variables at compile-time, so we create placeholders which we can then replace with the serialized values at run-time. 

---

Next up we combine these new assignments and the body, degensym it, put it in a block (if specified) and turn the code into a string:
```nim
# 2. Stringify code
var code = newStmtList(capAssignments, body).copyNimTree()
code.degensymAst()

if putCodeInBlock:
  code = newBlockStmt(code)
var codeText = code.toStrLit
```
So We will basically have this string in the end:
```nim
let codeText = """
let x = parseJson(placeholder`gensym1).to(int)
let y = parseJson(placeholder`gensym2).to(string)
echo x, y
"""
```

---

Now we have come to the most complex part of it all: generating the replacements of the placeholders with the serialized json. We start off with assigning the codeString to an actual variable:
```nim
let codeTextIdent = genSym(NimSymKind.nskVar, ident="codeText")
result = newStmtList()
result.add newVarStmt(codeTextIdent, codeText)
```
So this will generate the following code:
```nim
var codeText = """
let x = parseJson(placeholder`gensym1).to(int)
let y = parseJson(placeholder`gensym2).to(string)
echo x, y
"""
```
Then we loop over each of the capture variables:
```nim
for i in 0 .. captureVars.high:
```
and for each one we do the following:
```nim
# Get string literal of the placeholder (e.g. "placeholder`gensym1")
let placeholder = placeholders[i].repr.newLit
# Get the variable symbol
let varIdent = captureVars[i]
# Generate the call to serialize the variable
let serializedValue = quote do:
  $(toJson(`varIdent`))
```
So the serialized value is basically obtained like this:
```nim
$(toJson(x))
```
The last thing we do in the loop is to do the actual string replacement of the placeholder with the json:
```nim
result.add quote do:
  `codeTextIdent` = `codeTextIdent`.replace(`placeholder`, "\"\"\"" & `serializedValue` & "\"\"\"")
```
This basically generates this code:
```nim
codeText = codeText.replace("placeholder`gensym1", &"{$(toJson(x))}")
```
And that is basically it. This final code on the C-side is:
```nim
var codeText = """
let x = parseJson(placeholder`gensym1).to(int)
let y = parseJson(placeholder`gensym2).to(string)
echo x, y
"""
codeText = codeText.replace("placeholder`gensym1", &"{$(toJson(x))}")
codeText = codeText.replace("placeholder`gensym2", &"{$(toJson(y))}")
```

HugoGranstrom:
> the first example we show should not probably be one with a captured variable

Good point :+1: 

> we should mention the multiple apis and what are the sources of complexity, which should be essentially two: nim gensyms identifiers when compiling to js, we want to capture variables.

WIll add it to the `Internal Workings` section :+1: 

> regarding the multiple api the global js script api it seems to me is a novel and smart idea on how to solve the sources of complexity, so it should be given a bit more appreciation.

Do you have any concrete examples in mind? (I don't :rofl: )

> also on capturing variable it is now evident to me (especially looking at the examples of two counters) that it is important for passing ids and have reusable widgets that contain both rawHtml and Js code. At first I thought its main use case was to pass complex objects generated from C, but I guess that is an additional (cool) use case (on which I do not think I have really seen examples, I mean in principle we could train a neural network with arraymancer and translating to a easier structure where we could do the inference on js... you know just to start with simple ideas :D).

Haha yes, it has many use-cases :) I haven't tried it out with any complex datatypes yet tbh, might be a good excuse to test its limits and do something cool with it as you say :sunglasses: In your planned Machine learning tutorial on SciNim for example, if we trained some simple machine learning model and provided an interactive way of testing it using nbJs somehow. Basically training it in the C-side and making predictions on the JS side with user inputs.

HugoGranstrom:
> Ok, with the in-code-review and last comment, I think I have exhausted my review on current code. Hey I am taking this literally rofl:

Haha that's good! :smile: Didn't expect to sit nearly 2h answering :rofl: But that time will be well spent if it means I can save more hours in the future now that you know how this all works and can cover for me ;)

pietroppeter:
Thanks a lot for the answer and the great write up!

I have been reading and thinking about this stuff, and I have had some interesting thoughts (some outside of scope of this PR). Before doing that, I will like first to list stuff that is undecided, to do or needs more discussion:

from in-code review:
- [ ] first 4 remarks are action to be taken, small details, when we close them we resolve the conversation
- [ ] **decision to take**: do we interleave global scope and blocks or not?
- [ ] change logic of bump gensym, likely adding a counter and looping inside the macro. 👍 

rest of conversation.
- [ ] **decision to take**: do we rename `nbJsFromCodeGlobal` and `nbJsFromCode` to `nbJsFromCode` and `nbJsFromCodeInBlock` (agree that we should skip add)?
- [ ] **decision to take**: should we allow to add a final js block outside of nbSave?
- [ ] **decision to take**: should we do something about possibility of showing code in JsFromCode?
- [ ] move `compileNimToJs` from `renders.nim` to `jsutils.nim`
- [x] light on why we need the two stages, the cache table and degensym
  - the explanation looks very convincing to me, still need to understand details, will do it on my own time trying stuff when I am in the mood
- [x] explanation of variable captures
  - again here good explanation, thanks, will need to work more on it on my own
- [ ] first example will not be with capture variables
- [ ] will add a `internal working` sections on sources of complexity
- [x] not really have concrete example on how to show appreciation about single js script with global and block sections.

first off, did I miss something? I intentionally did not quote here, but I tried to go with order of appearance.

pietroppeter:
now onto more discussions on decisions to take.

### do we interleave global scope and blocks or not?

quoting you:

> order of execution changes. So it depends on what we want the global code to be able to do. I like the global-at-the-top approach because it allows us to do customizations of global variables defined by other blocks.

honestly on this I would be to keep the flow as is given by how the user added the blocks. It feels really odd to me to put them in another order than what appears in the source, I am not sure if the reordering might have unintended side effects (I cannot think of any at the moment). the use case we would not support maybe can be done in other ways.

### do we rename nbJsFromCodeGlobal and nbJsFromCode to nbJsFromCode and nbJsFromCodeInBlock?

> I think that the most recommended one (the block version) should have the shortest name, though...

I do disagree on this one too (sorry!). It seems to me the real essential one, the one that you for sure will use will be the global one. you will likely import stuff and I can imagine that most of the time you might have a single block of JsFromCode and nothing else. It can be that in library work (e.g. a widget), you need to use most often a InBlock one, but that's because you want to make sure not to have conflicts and stuff, so it is an expected change. I can understand that the block version is "safer" (and in this way it is more in line with how templates by default are not dirty), but naming the block version InBlock is also consistent with the name of nbCode and nbCodeInBlock.

### should we allow to add a final js block outside of nbSave?

on this one I have a new idea. it has to do with the fact that I do not like to much the fact that as they are now, the `nbJsFromCode` blocks are not really blocks (they do not add a NbBlock!). this prompted me to come up with a new type of block that I am calling the "collector" block. In this case we go back to have that `nbJsFromCode` (and `nbJsFromCodeInBlock) that add NbBlock (with command as in their names and we could use `code` and `output` as holders of `originalCode` and `transformedCode`). Those block will NOT have a render function (we might check if this generates a warning and find a way to turn it off. But we add a block at the end (`nbJsFromCodeCollect`? probably we might find a better name) that looks at the blocks in `NbDoc` (hey nothing forbids a block to look at the full doc when rendering! it has access to it!) and puts together the pieces of code in a single string. This would have other benefits: 1) we should not need anymore the additional fields in `NbDoc` (I think); 2) in principle we can change the behavior of the collect block to implement the alternative flow (all globals and then all in blocks); 3) also for showing source code of JsBlock we can do as we were doing before (and granted it might not make to much sense to show it, indeed we should not care too much about this).

What do you think of this option?
It leaves me incidentally without the last point to discuss (and the discussion on the first one is also defused).

pietroppeter:
finally, just to give an idea on something out of scope. I think I am realizing that we should probably refactor again the `NbBlock` and **embrace inheritance** and use methods. I never thought about it because I was kind of suspicious of oop approaches and I also have never used them but in Nim's context (and frankly, scarcely also elsewhere) the more I think about it, the more it makes perfect sense. extending or customizing a block should be just inheriting from it. the base block should have just the (new) `data: JsonNode` field (or possibly even nothing), and every block would have its natural fields (just nbCode will keep the `code` and `output` fields, the other would change). Having natural fields would mean we do not need to put stuff in a context/data object and have a bad api taking them out. Container blocks could be easier to implement. Even `NbDoc` could possibly inherit from `NbBlock` and we might have a container block for the head part of the document...

Ok, I might be pushing it to far, especially considering that I have very limited experience with method and stuff, but I got very excited today thinking about this (so be gentle to communicate to me if there are some evident reason why this would not work 😝). Anyway, this is not something decided or soon, I have to digest the idea, make experience with inheritance and methods in Nim (and the vegalite/etamin project I mentioned in science chat is likely a good place to experiment with that, it actually was likely the reason I am thinking about this stuff) and will open an issue when I am ready to give more explanation. It came up though while thinking about this PR and I cannot avoid giving a preview here.

pietroppeter:
And I do agree that it would be very cool to use interactive stuff in the ML tutorial!

HugoGranstrom:
> first off, did I miss something?

I think you got everything :+1: Nice to have it all listed out :D School is starting again tomorrow so we'll see when I'll manage to finish this :rofl: 

> honestly on this I would be to keep the flow as is given by how the user added the blocks. It feels really odd to me to put them in another order than what appears in the source, I am not sure if the reordering might have unintended side effects (I cannot think of any at the moment). the use case we would not support maybe can be done in other ways.

It's only odd if it's not what you expect from it. But I propose a compromise: we have 3 templates, `nbJsFromCode` (not in block, is interleaved with blocks), `nbJsFromCodeInBlock` (in block), `nbJsFromCodeGlobal` (not in block, placed at the top of the file). This way you get your naming scheme (which I agree with it the most consistent with our current naming of other blocks), and I get my global block. This way the user can decide for themselves how they want their code to be structured.  How does this sound? 

> It can be that in library work (e.g. a widget), you need to use most often a InBlock one, but that's because you want to make sure not to have conflicts and stuff, so it is an expected change. I can understand that the block version is "safer" (and in this way it is more in line with how templates by default are not dirty), but naming the block version InBlock is also consistent with the name of nbCode and nbCodeInBlock.

I agree with you on this one. I'm coming in from a very library-centric viewpoint, so good that you made me realize that :smile: 

> on this one I have a new idea. it has to do with the fact that I do not like to much the fact that as they are now, the nbJsFromCode blocks are not really blocks (they do not add a NbBlock!). this prompted me to come up with a new type of block that I am calling the "collector" block.

I actually like this idea :) Will try and implement it :+1: 
Point 1 and 3 are good. But point 2 is not a good idea imo. We shouldn't change the semantics of the nbJs blocks like that, that could really screw up the ordering of blocks. Especially if two different libraries want to use different orderings :sweat_smile: But if we go with the `nbJsFromCode/InBlock/Global` route, we wouldn't have to care about this.

HugoGranstrom:
> finally, just to give an idea on something out of scope. I think I am realizing that we should probably refactor again the NbBlock and embrace inheritance and use methods. I never thought about it because I was kind of suspicious of oop approaches and I also have never used them but in Nim's context (and frankly, scarcely also elsewhere) the more I think about it, the more it makes perfect sense. extending or customizing a block should be just inheriting from it.

That makes a lot of sense now that you mention it :o The only problematic thing I can come up with is if you wanted to inherit multiple blocks. For example if we want to inherit from both `nbText` and `nbCode`. I don't think Nim supports multiple inheritance. Or would we keep the `renderProc/Plan` as well? 

I'm not that experience in OOP either tbh, but if we could make it work and make it easy to extend blocks, then it's definetly a route worth to investigate :)

pietroppeter:
>  I propose a compromise: we have 3 templates, nbJsFromCode (not in block, is interleaved with blocks), nbJsFromCodeInBlock (in block), nbJsFromCodeGlobal (not in block, placed at the top of the file) ... How does this sound?

sounds great, I like the idea! indeed it also solves the bad design of having 

I think we are mostly aligned now! 🚀 when you start working on this you could take the bullet list and update the initial thread.

>  School is starting again tomorrow so we'll see when I'll manage to finish this 🤣

take your time! 😃 

> >  I am realizing that we should probably refactor again the NbBlock and embrace inheritance
> That makes a lot of sense now that you mention it :o 

right? happy that I am not completely off the mark! 😅 

> The only problematic thing I can come up with is if you wanted to inherit multiple blocks. For example if we want to inherit from both nbText and nbCode. I don't think Nim supports multiple inheritance. Or would we keep the renderProc/Plan as well?

yep, there are plenty of stuff we need to find out on this route. on multiple inheritance there is an interesting [RFC thread](https://github.com/nim-lang/RFCs/issues/262) (in particular macro approaches would be possible). Anyway even with single inheritance situation seems better than what we have now.

HugoGranstrom:
> sounds great, I like the idea! indeed it also solves the bad design of having
> I think we are mostly aligned now! rocket when you start working on this you could take the bullet list and update the initial thread.

Nice! :D Will do! 

> take your time! smiley

Thanks :) I'll aim at having it done next weekend...

> right? happy that I am not completely off the mark! sweat_smile
> yep, there are plenty of stuff we need to find out on this route. on multiple inheritance there is an interesting https://github.com/nim-lang/RFCs/issues/262 (in particular macro approaches would be possible). Anyway even with single inheritance situation seems better than what we have now.

Haha, you might definitely be on to something. A superweird, but kinda related, idea: using ECS and Polymorph instead of inheritance :rofl: It's not a serious proposal, but it would be a good exercise if one wants to learn Polymorph for something that isn't game-related. To implement a nimib-like API on top of it, just for fun.

pietroppeter:
> A superweird, but kinda related, idea: using ECS and Polymorph instead of inheritance 🤣

haha, thought about that also when I read the thread. honestly I know even less about ECS than OOP so it might be for the next iteration. I do have in mind to play with polymorph though (e.g. https://github.com/pietroppeter/nblog/issues/13), it really looks great!

HugoGranstrom:
> haha, thought about that also when I read the thread. honestly I know even less about ECS than OOP so it might be for the next iteration. I do have in mind to play with polymorph though (e.g. https://github.com/pietroppeter/nblog/issues/13), it really looks great!

Yes that seems like a more reasonable thing to start with :)

HugoGranstrom:
The refactoring of the code was easier than I expected. The collector proc was really convenient :D Now I just have to update the docs with better examples and more explanations.

HugoGranstrom:
I have a draft of the new `interactivity` document now. Would appreciate a review of it as I'm feeling it is a bit too technical in the first part. But I don't see a good way of explaining it without mentioning the nim common file all blocks will be put inside

docsrc/interactivity.nim
```diff
@@ -2,57 +2,125 @@ import nimib
...
+nb.renderPlans["nimibCode"] = nb.renderPlans["nbCode"]
+
 nbText: hlMd"""
-# Creating interactive components in Nimib
+# Creating interactive components in nimib
 
 Nimib can easily be used to create static content with `nbText` and `nbCode`, but did you know that you can create interactive
 content as well? And that you can do it all in Nim even! This can be achieved using either the `nbJsFromCode`-API or `nbKaraxCode`.
 They work by compiling Nim code into javascript and adding it to the resulting HTML file.
 This means that arbitrary Javascript can be written but also that Karax, which compiles to javascript, also can be used.
```
pietroppeter, line 19: did not notice last time but "also" is repeated, we could omit the second also

docsrc/interactivity.nim
```diff
@@ -2,57 +2,125 @@ import nimib
 
 nbInit
 
+template nimibCode*(body: untyped) =
```
pietroppeter, line 5: indeed nimibCode is useful. should we add it do nimib? I do not mind the off-convention naming (nimibCode instead of nbCode), since it is about nimib itself. It could actually start a new convection that this kind of blocks (honestly not sure if there will be others), could be all start with nimib.

docsrc/interactivity.nim
```diff
@@ -2,57 +2,125 @@ import nimib
...
+
 nbText: hlMd"""
-# Creating interactive components in Nimib
+# Creating interactive components in nimib
 
 Nimib can easily be used to create static content with `nbText` and `nbCode`, but did you know that you can create interactive
 content as well? And that you can do it all in Nim even! This can be achieved using either the `nbJsFromCode`-API or `nbKaraxCode`.
 They work by compiling Nim code into javascript and adding it to the resulting HTML file.
 This means that arbitrary Javascript can be written but also that Karax, which compiles to javascript, also can be used.
+The available template are:
```
pietroppeter, line 20: I would actually start with the "by default we compile all into a single js file". 
A start could be:

```md
In the same way that code from nbCode blocks are all compiled into a single file,
all code to be compiled in javascript will be put in a single file. 
This has the advantage that a single compilation is performed
and code from a previous block can be used in subsequent blocks.
The api looks like this:

- `nbJsFromCode`: nim code will be appended to the file and compiled during `nbSave`.
- `nbJsFromCodeInBlock`: same as `nbJsFromCode` but the code is put inside a `block`.
- `nbJsFromCodeGlobal`: the code here will be put at the top of the file inside the global code.

If you wish to compile to a separate file you can do that.
Indeed this is what is done for a special block that allows you to use karax removing all boilerplate:

- `nbJsFromCodeOwnFile`: compile to js as its own file.
- `nbKaraxCode`: Sugar on top of `nbJsFromCodeOwnFile` for writing Karax components.
```

docsrc/interactivity.nim
```diff
@@ -2,57 +2,125 @@ import nimib
...
+nimibCode:
+  nbJsFromCode:
+    let x = "Hello world!"
+    echo x
 
+nbText: hlMd"""
+If you now go to your browser's javascript console you should see `Hello world` printed there.
+So the code we passed to `nbJsFromCode` has been compiled to Javascript and is run by your browser!
 
+### Capturing variables
```
pietroppeter, line 45: at the end of capturing variables section we could point out that capturing variables is essential for a js widget library and refer to counter document to show an example of that.

docsrc/interactivity.nim
```diff
@@ -2,57 +2,125 @@ import nimib
...
+    var globalVar = 1
+  nbJsFromCode:
+    echo "First block: ", globalVar
+    globalVar += 1
+  nbJsFromCode:
+    echo "Second block: ", globalVar
+
+nbText: hlMd"""
+## nbJsFromCodeOwnFile
+The above-mentioned nbJs blocks are all compiled in the same file. But if you want to comile a code block
```
pietroppeter, line 113: missing a P in comPile

pietroppeter (commented):
I had to attend other stuff in the middle of the review, will complete later

docsrc/interactivity.nim
HugoGranstrom: I don't think it matters that much that it doesn't start with `nb` as it is only useful when showing code blocks with use nimib templates. And it will most likely be you and me who are using it.  So I'm for adding it :+1:  should I add it in this PR or do you want it in a separate one?

docsrc/interactivity.nim
HugoGranstrom: Great idea, thanks :D Rephrased it in a few places

docsrc/interactivity.nim
pietroppeter: This one is fine I guess

src/nimib.nim
```diff
@@ -132,22 +137,9 @@ template nbRawHtml*(content: string) =
 template nbJsFromStringInit*(body: string): NbBlock =
   var result = NbBlock(command: "nbCodeToJs", code: body, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```
pietroppeter, line 138: should we change name of the partial from `nbCodeToJs` to `nbCodeFromJs`?
```suggestion
  var result = NbBlock(command: "nbCodeFromJs", code: body, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```

src/nimib/renders.nim
```diff
@@ -48,7 +33,8 @@ proc useHtmlBackend*(doc: var NbDoc) =
 <pre><code class="{{ext}} hljs">{{content}}</code></pre>
 """
 
-  doc.partials["nbCodeToJs"] = """
+  doc.partials["nbCodeToJs"] = "{{>nbJsScriptNimSource}}" # the script is handled by collector block
```
pietroppeter, line 36: also here:
```suggestion
  doc.partials["nbCodeFromJs"] = "{{>nbJsScriptNimSource}}" # the script is handled by collector block
```

src/nimib/renders.nim
```diff
@@ -48,7 +33,8 @@ proc useHtmlBackend*(doc: var NbDoc) =
...
+  doc.partials["nbCodeToJsOwnFile"] = """
```
pietroppeter, line 37:
```suggestion
  doc.partials["nbCodeFromJsOwnFile"] = """
```

src/nimib.nim
```diff
@@ -161,8 +153,31 @@ template nbJsFromString*(body: string) =
   script.addToDocAsJs
 
 template nbJsFromCode*(args: varargs[untyped]) =
-  let script = nbJsFromCodeInit(args)
-  script.addToDocAsJs
+  let (code, originalCode) = nimToJsString(putCodeInBlock=false, args)
+  var result = NbBlock(command: "nbCodeToJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```
pietroppeter, line 157:
```suggestion
  var result = NbBlock(command: "nbCodeFromJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```

src/nimib.nim
```diff
@@ -161,8 +153,31 @@ template nbJsFromString*(body: string) =
...
+  result.context["transformedCode"] = code
+  result.context["putAtTop"] = false
+  result.addToDocAsJs
+
+template nbJsFromCodeInBlock*(args: varargs[untyped]) =
+  let (code, originalCode) = nimToJsString(putCodeInBlock=true, args)
+  var result = NbBlock(command: "nbCodeToJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```
pietroppeter, line 164:
```suggestion
  var result = NbBlock(command: "nbCodeFromJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```

src/nimib.nim
```diff
@@ -161,8 +153,31 @@ template nbJsFromString*(body: string) =
...
+  result.context["transformedCode"] = code
+  result.context["putAtTop"] = false
+  result.addToDocAsJs
+
+template nbJsFromCodeGlobal*(args: varargs[untyped]) =
+  let (code, originalCode) = nimToJsString(putCodeInBlock=false, args)
+  var result = NbBlock(command: "nbCodeToJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```
pietroppeter, line 171:
```suggestion
  var result = NbBlock(command: "nbCodeFromJs", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```

src/nimib.nim
```diff
@@ -161,8 +153,31 @@ template nbJsFromString*(body: string) =
...
+  result.context["transformedCode"] = code
+  result.context["putAtTop"] = true
+  result.addToDocAsJs
+
+template nbJsFromCodeOwnFile*(args: varargs[untyped]) =
+  let (code, originalCode) = nimToJsString(putCodeInBlock=false, args)
+  var result = NbBlock(command: "nbCodeToJsOwnFile", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```
pietroppeter, line 178:
```suggestion
  var result = NbBlock(command: "nbCodeFromJsOwnFile", code: originalCode, context: newContext(searchDirs = @[], partials = nb.partials), output: "")
```

src/nimib/renders.nim
```diff
@@ -58,7 +44,9 @@ proc useHtmlBackend*(doc: var NbDoc) =
   doc.renderPlans = initTable[string, seq[string]]()
   doc.renderPlans["nbText"] = @["mdOutputToHtml"]
   doc.renderPlans["nbCode"] = @["highlightCode"] # default partial automatically escapes output (code is escaped when highlighting)
-  doc.renderPlans["nbCodeToJs"] = @["compileNimToJs", "highlightCode"]
+  doc.renderPlans["nbCodeToJsOwnFile"] = @["compileNimToJs", "highlightCode"]
```
pietroppeter, line 47:
```suggestion
  doc.renderPlans["nbCodeFromJsOwnFile"] = @["compileNimToJs", "highlightCode"]
```

src/nimib/renders.nim
```diff
@@ -58,7 +44,9 @@ proc useHtmlBackend*(doc: var NbDoc) =
...
+  doc.renderPlans["nbCodeToJs"] = @["highlightCode"]
```
pietroppeter, line 48:
```suggestion
  doc.renderPlans["nbCodeFromJs"] = @["highlightCode"]
```

pietroppeter (changes requested):
looks good. just add a few suggestion to align block names to new naming convention.
I may have lost some `nbJsToCode` along the way (especially if they do not appear in the diff..).

Will now look again at the preview docs but I guess we are basically done! :)

pietroppeter:
Had a look at the preview docs and had a few more remarks:
- first JsFromCode code is not shown in the document but it should
- we could mention at the end of nbKaraxCode section of interactivity.nim that we have another example of using nbKaraxCode in caesar
  - in index.nim short description of caesar we refer to it as being built with nbJsFromCode while it uses nbKaraxCode

I think with this I finally completed the review. Sorry for the delay and again great work on this!

src/nimib.nim
HugoGranstrom: Yes that a good idea. I've been afraid to do it because I would miss to change it somewhere :rofl:

HugoGranstrom: Shouldn't it be `"nbJsFromCode"` though? :sweat_smile:

HugoGranstrom:
> first JsFromCode code is not shown in the document but it should

Not sure I understand what you mean :/

> we could mention at the end of nbKaraxCode section of interactivity.nim that we have another example of using nbKaraxCode in caesar
> in index.nim short description of caesar we refer to it as being built with nbJsFromCode while it uses nbKaraxCode

Will fix!

HugoGranstrom:
> first JsFromCode code is not shown in the document but it should

Now I found it. Changed `nbCode` to `nimibCode` and it worked.

> I think with this I finally completed the review. Sorry for the delay and again great work on this!

:partying_face: 
No worries :smile: Hopefully I shouldn't have to update the nbJs stuff in  a while now that it's so simple, so the question is what I should do now :rofl:

pietroppeter:
> the question is what I should do now

well, there is plenty of open issues... 😆

pietroppeter (approved):

pietroppeter:
ready to merge, right?
I guess after merging we should also bump version and release, right?
if you want to make the preparations (bump version, finalize changelog for release, ...) in this PR, otherwise I'll do it later

HugoGranstrom:
🤣 very true, 51 open issues should have something yummy in them. 

I can bump the version and changelog when I get home 👍 I think this is my first time making a new release on my own? 😯

pietroppeter:
> I think this is my first time making a new release on my own? 😯

then go ahead and enjoy the release! 🚀

HugoGranstrom:
Versions are bumped, hope I didn't forget anything now...
Will have a look through the preview before merging this.

pietroppeter:
ah, maybe add a line about the new `nimibCode` in changelog?

HugoGranstrom:
Good point :+1 Looked through the docs before pushing the last commit and everything seems to work fine. So I'll merge this after dinner unless we find something else until then.

pietroppeter:
looks good to me.

the more I think about it the more stuff it pops to mind. we probably should link the NimConf 2022 presentation (and slides) in the readme/index? not sure in the minimal way we did for 2021 or showing the video thumbnail (would look more attractive). to be honest this could all be separate PRs but since we are here...

HugoGranstrom:
It would be nice with a thumbnail, but we'd just have to get a better thumbnail where we both are in it. I have uploaded one such example. But I think it is better if we merge this first so we can get a static link to it in the main branch instead before linking to it. So I'll merge this and create a follow-up PR for that.

PR #152, opened 2022-11-13
add nimconf to readme and index

HugoGranstrom:
let's try again...


HugoGranstrom:
This looked much better, now the question is why we got an error in the `docs` Github task but not here. Maybe it was just bad luck and it would work if we tried again :shrug:

PR #158, opened 2022-11-28
add nbCodeDisplay and nbCodeAnd

### LINKED ISSUE #154: Code shown with `nbJsShowSource` is not as in source (but comes from ast)
I am using `nbJsShowSource` for showing p5js examples (see e.g. https://pietroppeter.github.io/nim-p5/polygons.html).
The code comes from ast instead of being read from source, with all classical problems (no comments and weird indents).

pietroppeter:
fix #154, instead of creating a `nbJsFromCodeDisplay` as suggested there (I would need also the Global, InBlock and possibly other variations), I only add this template. It will not work with templates that have more than one body argument (so it will not work directly with the Js templates when capturing variables, but you can wrap that call in a template with single body argument). Seems an easy enough change. Already using it in p5nim, choosing to go with nbCode after the template call (inverse would have been fine too).

I am also adding a `nbCodeAnd` to execute code in c backend **and** js backend (when applied to `nbJsFromCode`).
Both are documented in `interactivity.nim` because it makes sense to mention them there but they are generic.

I have already started using `nbCodeDisplay` in p5nim (see https://github.com/pietroppeter/p5nim/pull/9) and I plan to use `nbCodeAnd` in adventofnim (to solve a puzzle and use the same structure to visualize it with p5).

- [x] add nbCodeDisplay template
- [x] add nbCodeDisplay template
- [x] use it and document it (in interactivity.nim)
- [x] add tests for both
- [x] line in changelog
- [x] bump to 0.3.4

### DIFF
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -197,6 +197,11 @@ template nbJsShowSource*(message: string = "") =
 template nbCodeToJsShowSource*(message: string = "") {.deprecated: "Use nbJsShowSource instead".} =
   nbJsShowSource(message)
 
+template nbCodeDisplay*(tmplCall: untyped, body: untyped) =
+  newNbCodeBlock("nbCode", body):
+    discard
+  tmplCall:
+    body
 
 template nbClearOutput*() =
   if not nb.blk.isNil:
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -189,12 +189,12 @@ when moduleAvailable(karax/kbase):
     nbRawHtml: "<div id=\"" & rootId & "\"></div>"
     nbKaraxCodeBackend(rootId, args)
 
-template nbJsShowSource*(message: string = "") =
+template nbJsShowSource*(message: string = "") {.deprecated: "Use nbCodeDisplay instead".} =
   nb.blk.context["js_show_nim_source"] = true
   if message.len > 0:
     nb.blk.context["js_show_nim_source_message"] = message
 
-template nbCodeToJsShowSource*(message: string = "") {.deprecated: "Use nbJsShowSource instead".} =
+template nbCodeToJsShowSource*(message: string = "") {.deprecated: "Use nbCodeDisplay instead".} =
   nbJsShowSource(message)
 
 template nbCodeDisplay*(tmplCall: untyped, body: untyped) =
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -198,8 +198,16 @@ template nbCodeToJsShowSource*(message: string = "") {.deprecated: "Use nbJsShow
   nbJsShowSource(message)
 
 template nbCodeDisplay*(tmplCall: untyped, body: untyped) =
+  ## display codes used in a template (e.g. nbJsFromCode) after the template call
+  tmplCall:
+    body
   newNbCodeBlock("nbCode", body):
     discard
+
+template nbCodeAnd*(tmplCall: untyped, body: untyped) =
+  ## can be used to run code both in c and js backends (e.g. nbCodeAnd(nbJsFromCode))
+  nbCode: # this should work because template name starts with nbCode
+    body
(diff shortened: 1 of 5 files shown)


pietroppeter:
hi @HugoGranstrom, if you want to give a quick review I think these two templates can be useful and they are a nice addition. After this is merged I will tag a new release (changelog and nimble version already updated).

src/nimib.nim
```diff
@@ -189,14 +189,27 @@ when moduleAvailable(karax/kbase):
...
+template nbCodeDisplay*(tmplCall: untyped, body: untyped) =
+  ## display codes used in a template (e.g. nbJsFromCode) after the template call
+  tmplCall:
+    body
+  newNbCodeBlock("nbCode", body):
+    discard
+
+template nbCodeAnd*(tmplCall: untyped, body: untyped) =
+  ## can be used to run code both in c and js backends (e.g. nbCodeAnd(nbJsFromCode))
+  nbCode: # this should work because template name starts with nbCode
```
HugoGranstrom, line 209: It's nice that we are embracing this lucky bug at least :rofl:

src/nimib.nim
```diff
@@ -189,14 +189,27 @@ when moduleAvailable(karax/kbase):
...
   nbJsShowSource(message)
 
+template nbCodeDisplay*(tmplCall: untyped, body: untyped) =
+  ## display codes used in a template (e.g. nbJsFromCode) after the template call
+  tmplCall:
+    body
+  newNbCodeBlock("nbCode", body):
+    discard
+
+template nbCodeAnd*(tmplCall: untyped, body: untyped) =
```
HugoGranstrom, line 207: What if the user wants to run the code in another block than `nbCode`, eg. `nbCodeInBlock`? I think it would make sense that `nbCodeAnd` takes in two `tmplCall` so the user can choose for themselves. Plus, it becomes more general. I don't know how this works out with nbCodeAsInSource though :sweat_smile: So maybe it isn't possible :shrug:

HugoGranstrom (commented):
I like the idea, but I don't really see how it could be used for multiple inputs to the template. Could you give a minimal example of how one would capture a variable using `nbCodeDisplay`? 

Asked a question on the possibility to extend `nbCodeAnd` to arbitrary blocks for both.

pietroppeter:
> Could you give a minimal example of how one would capture a variable using nbCodeDisplay?

original code:
```nim
nbJsFromSource(captureA, captureB):
  discard captureA
  discard captureB
```

transformed code:
```nim
template myNbJsWithCaptures(body: untyped) =
  nbJsFromSource(captureA, captureB):
    body

nbCodeDisplay(myNbJsWithCaptures):
  discard captureA
  discard captureB
```

have not tried but I think it should work.

src/nimib.nim
pietroppeter: for this I think the quicker solution would be to provide with a `nbCodeInBlockAnd`.

In principle we could add a second template and do `nbThisAndThat(nbCodeInBlock, nbJsFromSource)` but this would not compose with `codeAsInsource` (which is actually the full purpose here).

src/nimib.nim
pietroppeter: yeah, maybe too much... still hopeful that at some point we will have a better solution for a composable CodeAsInSource (not working on it, just hope 😄 )

HugoGranstrom: Yeah, I'll try and think about it from time to time to see if I can get any strokes of genius. It's starting to bug me that it doesn't compose well. There's too much jumping through hoops to reuse a template 😅 So this might be my next issue to solve...

HugoGranstrom:
Everything looks good, you can merge when you're ready 👌😄

pietroppeter:
Thanks for the feedback. Good to have the view from outside!

PR #169, opened 2023-02-01
Adding devel to CI, fix #161, fix #149

### LINKED ISSUE #149: add devel to CI (and mulitple OSes)
- since nimib will be included in important packages (see https://github.com/nim-lang/Nim/pull/20697), we should add devel to CI.
- this is especially important for upcoming nim v2 (and after v2 is out, we should keep supporting 1.6)
- maybe we could also add windows and macos to CI.
- I think there are multiple clean CI setups around that use a matrix with multiple OSes and versions, we could copy one of those.

### LINKED ISSUE #161: nim v2.0 compatibility
have not tested yet how it works, but we should (and we should also add devel to CI).

as far as I know we do have an outstanding issue there, indeed when testing [adding nimib to important packages](https://github.com/nim-lang/Nim/pull/20697), an issue was [found related to toml dependency](https://github.com/status-im/nim-serialization/issues/45). The workaround currently is to revert using old gc (i.e. add `-mm:refc`).

Ideally we could:
- help fix the toml dependency (better option, I do not have ideas on how to proceed there currently)
- find a workaround in nimib (worse option and possibly not even something that could be done...)

pietroppeter:
I am adding both stable (that will become 2.x soon) and devel (which is candidate for 2.x). Also update versions of cache and setup-nim-action.

Done to check if recent toml fix makes nimib works in 2.0rc

We could also decide to expand the test matrix to multiple os.

fix #161 and #149

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,11 +7,18 @@ on:
 jobs:
   tests:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        nim:
+          - '1.6.x'
+          - 'stable'
+          - 'devel'
+    name: Nim ${{ matrix.nim }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: jiro4989/setup-nim-action@v1.3.15
+      - uses: actions/checkout@v3
+      - uses: jiro4989/setup-nim-action@v1.4.3
         with:
-          nim-version: '1.6.x'
+          nim-version: ${{ matrix.nim }}
       - run: nimble -y install
       - run: nimble docsdeps
       - run: nimble test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
           - '1.6.x'
           - 'stable'
           - 'devel'
+      fail-fast: false
     name: Nim ${{ matrix.nim }}
     steps:
       - uses: actions/checkout@v3
--- a/nimib.nimble
+++ b/nimib.nimble
@@ -12,7 +12,8 @@ requires "nim >= 1.4.0"
 requires "tempfile >= 0.1.6"
 requires "markdown >= 0.8.1"
 requires "mustache >= 0.2.1"
-requires "toml_serialization >= 0.2.0"
+requires "parsetoml >= 0.7.0"
+requires "jsony >= 1.1.5"
 
 task docsdeps, "install dependendencies required for doc building":
   exec "nimble -y install ggplotnim@0.5.3 numericalnim@0.6.1 nimoji nimpy karax@1.2.2"
--- a/nimib.nimble
+++ b/nimib.nimble
@@ -1,6 +1,6 @@
 # Package
 
-version       = "0.3.5"
+version       = "0.3.6"
 author        = "Pietro Peterlongo & Hugo Granström"
 description   = "nimib 🐳 - nim 👑 driven ⛵ publishing ✍"
 license       = "MIT"
--- a/src/nimib/config.nim
+++ b/src/nimib/config.nim
@@ -1,4 +1,4 @@
-import types, os, toml_serialization
+import types, parsetoml, jsony, std / [json, os, math, sequtils]
 
 proc hasCfg*(doc: var NbDoc): bool = doc.cfgDir.string != ""
 
@@ -13,6 +13,64 @@ proc optOverride*(doc: var NbDoc) =
   if doc.options.homeDir != "":
     doc.cfg.homeDir = doc.options.homeDir
 
+
+proc customToJson*(table: parsetoml.TomlTableRef): JsonNode
+
+proc customToJson*(value: parsetoml.TomlValueRef): JsonNode =
+  case value.kind:
+    of TomlValueKind.Int:
+      %* value.intVal
+    of TomlValueKind.Float:
+      if classify(value.floatVal) == fcNan:
+        if value.forcedSign != Pos:
+          %* value.floatVal
+        else:
+          %* value.floatVal
+      else:
+        %* value.floatVal
+    of TomlValueKind.Bool:
+      %* $value.boolVal
+    of TomlValueKind.Datetime:
+      if value.dateTimeVal.shift == false:
+        %* value.dateTimeVal
+      else:
+        %* value.dateTimeVal
+    of TomlValueKind.Date:
+      %* value.dateVal
+    of TomlValueKind.Time:
+      %* value.timeVal
+    of TomlValueKind.String:
+      %* value.stringVal
+    of TomlValueKind.Array:
+      if value.arrayVal.len == 0:
+        when defined(newtestsuite):
+          %[]
+        else:
+          %* []
+      elif value.arrayVal[0].kind == TomlValueKind.Table:
+        %value.arrayVal.map(customToJson)
+      else:
+        when defined(newtestsuite):
+          %*value.arrayVal.map(customToJson)
+        else:
+          %* value.arrayVal.map(customToJson)
+    of TomlValueKind.Table:
+      value.tableVal.customToJson()
+    of TomlValueKind.None:
+      %*{"type": "ERROR"}
+
+proc customToJson*(table: parsetoml.TomlTableRef): JsonNode =
+  result = newJObject()
+  for key, value in pairs(table):
+    result[key] = value.customToJson
+
+
+proc loadTomlSection*[T](content, section: string, _: typedesc[T]): T =
+  let toml = parsetoml.parseString(content)
+  result = T()
+  if section in toml:
+    result = ($toml[section].customToJson()).fromJson(T)
+
 proc loadNimibCfg*(cfgName: string): tuple[found: bool, dir: AbsoluteDir, raw: string, nb: NbConfig] =
   for dir in parentDirs(getCurrentDir()):
     if fileExists(dir / cfgName):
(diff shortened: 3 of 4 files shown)


pietroppeter:
ok, now we have failure only in devel

ringabout:
It seems to be a regression of ORC.

ringabout:
```nim
macro getCodeAsInSource*(source: string, command: static string, body: untyped): string =
  ## Returns string for the code in body from source. 
  # substitute for `toStr` in blocks.nim
  let startPos = startPos(body)
  let filename = startPos.filename.newLit
  let endPos = finishPos(body)
  let endFilename = endPos.filename.newLit

  result = quote do:
    if `filename` notin nb.sourceFiles:
      nb.sourceFiles[`filename`] = readFile(`filename`)

    doAssert `endFilename` == `filename`, """
    Code from two different files were found in the same nbCode!
    If you want to mix code from different files in nbCode, use -d:nimibCodeFromAst instead. 
    If you are not mixing code from different files, please open an issue on nimib's Github with a minimal reproducible example."""

    getCodeBlock(nb.sourceFiles[`filename`], `command`, `startPos`, `endPos`)
```

It seems that in `nimib/sources`, the `filename` isn't semchecked properly, since it doesn't have a type at codegen phase.

HugoGranstrom:
Isn't this just a bug in the compiler and not a problem on our side?

ringabout:
Yeah, It's a weird compiler bug, though. I don't have a solution or a workaround.

HugoGranstrom:
Ok, that's problematic 🤔 because we do need the filename and it must come from the macro. I'll have a look in the coming days and see if I can find an alternative way 👍

ringabout:
You might report the bug to Nim if you happen to find a reduced case.

HugoGranstrom:
I'll give it a try 👍

HugoGranstrom:
I tried using nimib `0.3.0` and while it compiles it seems to have other problems with `orc`:
```
Traceback (most recent call last)
/home/hugo/.nimble/pkgs/nimib-0.3.0/nimib.nim(43) x_nim2
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/pure/os.nim(234) /
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/pure/os.nim(175) joinPath
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/pure/os.nim(143) joinPathImpl
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/pure/pathnorm.nim(98) addNormalizePath
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system.nim(3071) substr
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(908) alloc0
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(905) alloc
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(735) rawAlloc
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
Segmentation fault (core dumped)
```
So even if we manage to solve this specific issue I fear we have a few other orc bugs as well that isn't related to macros :upside_down_face:

HugoGranstrom:
But for the macro, it seems like it is the `startPos` and `endPos` that it doesn't like. The `filename` works totally fine on its own.

HugoGranstrom:
It seems to work if we create a literal of the `Pos` objects and pass them instead:
```
let startPosLit = startPos.newLit
let endPosLit = endPos.newLit
```

Then the question is whether the code without converting to a literal should work or not.

HugoGranstrom:
@pietroppeter This is the new `getCodeAsInSource`:
```
macro getCodeAsInSource*(source: string, command: static string, body: untyped): string =
  ## Returns string for the code in body from source. 
  # substitute for `toStr` in blocks.nim
  let startPos = startPos(body)
  let filename = startPos.filename.newLit
  let endPos = finishPos(body)
  let endFilename = endPos.filename.newLit

  let endPosLit = endPos.newLit
  let startPosLit = startPos.newLit

  result = quote do:
    if `filename` notin nb.sourceFiles:
      nb.sourceFiles[`filename`] = readFile(`filename`)

    doAssert `endFilename` == `filename`, """
    Code from two different files were found in the same nbCode!
    If you want to mix code from different files in nbCode, use -d:nimibCodeFromAst instead. 
    If you are not mixing code from different files, please open an issue on nimib's Github with a minimal reproducible example."""

    getCodeBlock(nb.sourceFiles[`filename`], `command`, `startPosLit`, `endPosLit`)
```

Now I'm getting the same runtime error as I got in nimib 0.3.0 though...

HugoGranstrom:
Issue for macro opened: https://github.com/nim-lang/Nim/issues/21326

HugoGranstrom:
I'm not able to wrap my head around the runtime error :disappointed: This reproduces it locally for me (I have installed `#head` of both `toml_serialization` and `serialization`):
```nim
import os
import "$nim/compiler/pathutils"

import nimib, nimib/[options, config]

var nb: NbDoc
nb.initDir = getCurrentDir().AbsoluteDir
loadOptions nb
loadCfg nb

let thisFile = "/home/hugo/code/nim/nimib/x_pathutils.nim".AbsoluteFile

echo thisFile
echo thisFile.splitFile
echo thisFile.splitFile.dir
echo thisFile.splitFile.dir.string / "a"
```

This gives this output:
```
[nimib] config file found: /home/hugo/code/nim/nimib/nimib.toml
/home/hugo/code/nim/nimib/x_pathutils.nim
(dir: /home/hugo/code/nim/nimib, name: "(dir: \x00\x00\x00ls", ext: ".nim")
/home/hugo/code/nim/nimib
Traceback (most recent call last)
/home/hugo/code/nim/nimib/x_pathutils.nim(16) x_pathutils
/home/hugo/.choosenim/toolchains/nim-1.6.10/compiler/pathutils.nim(30) splitFile
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/pure/os.nim(689) splitFile
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system.nim(3071) substr
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(908) alloc0
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(905) alloc
/home/hugo/.choosenim/toolchains/nim-1.6.10/lib/system/alloc.nim(735) rawAlloc
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
Segmentation fault (core dumped)
Error: execution of an external program failed: '/home/hugo/.cache/nim/x_pathutils_d/x_pathutils_7DB931EC51DD8FF47DEDE3914D845AF3C1979994 '
```

And we can note several things. First, the `name` field of the `splitFile` looks really weird, but when printed on its own it looks ok. Second thing is that we are loading a `nimib.toml` file. If I remove it (or remove `loadOptions` or `loadCfg`) the program runs just fine. How could the toml file possibly affect the variable `thisFile` which has nothing to do with it?!?! 

I have no clue what's going on and what to blame tbh. Is it toml_serialization with yet another orc bug or a bug in the compiler which just happens to show up when using toml_serialization for a totally different part of the program?

pietroppeter:
This is getting really weird, did not have too much time to look into this myself and not sure how to help at the moment!

pietroppeter:
I think anyway you did make a few good steps toward resolution, thanks for that! Last example with toml vs compiler bug I guess it could be nice to have it further simplified but it does not seem to be easy.

HugoGranstrom:
It's definitely a toml problem, managed to reduce it to this:
```nim
import toml_serialization

type NbConfig* = object
  srcDir*, homeDir*: string

let f = readFile("nimib.toml")

let t = Toml.decode(f,  NbConfig, "nimib")

echo t # error here
```
```toml
# nimib.toml
[nimib]
srcDir = "docsrc"
homeDir = "docs"
```

So we are still blocked by toml_serialization. I'm honestly tempted to change toml library to https://github.com/NimParsers/parsetoml which I just tried and it works under `orc`. We don't need any fancy bells and whistles so using Status' libraries is overkill in my opinion. I might even go as far as to say that we should avoid using their libraries at all if we want our code to work on the latest versions of Nim in the future. Having to wait months for them to fix things is not viable... Then I prefer to use simpler libraries that don't try to do anything fancy with threads and pointers, they are less prone to break.

pietroppeter:
Ok then we should report the error on toml serialization, that at least is good news since there is not anything weirder. I agree we should consider changing library, apart the various dependencies we get from status (which I never liked), it makes it hard to fix it ourselves. The api of parseToml though at the moment does not support how we use it in nimib (also nimibook will be affected). The advantage of toml  serialization was that you could target parsing a specific section and parse using a type as specification, similar to what jsony does. It is possible we might be able to add that feature to parseToml, I will look into it.

HugoGranstrom:
`parsetoml` has an open [issue](https://github.com/NimParsers/parsetoml/issues/59) on parsing directly into an object. And a workaround is going through json (toml → json → nim). And as for the section, we just have to create a wrapper proc which checks for a `nimib` section in the file before trying to read from the section. So there isn't really any features that are missing, we just have to restructure our code a bit. 

Have opened an issue on toml_serialization: https://github.com/status-im/nim-toml-serialization/issues/62

pietroppeter:
Do you want to give it a try?

HugoGranstrom:
I don't have much free time this weekend but I could try it next weekend 👍

pietroppeter:
should we bump to 0.4 or a minor bump to 0.3.6 given the parsetoml change? I would go with 0.3.6 but fine with the other too, @HugoGranstrom?

pietroppeter:
and for the changelog I was thinking of bumping, releasing curating the automatic changelog and adding the results to changelog.md. We could actually change our workflow to "not update changelog during PRs" and only make sure that message in PR merge commit is a good for the automatic changelog provided by Github. I would still want for the moment to maintain the changelog.md (especially for the history part) by updating it after the release.

pietroppeter:
ready to merge and release if you are ok with bumping to 0.3.6 @HugoGranstrom

HugoGranstrom:
0.3.6 sounds good to me! 👌 Release it! 

Changelog idea sounds good for the future 👍

PR #171, opened 2023-02-03
Check ci

HugoGranstrom:
I just want to check the CI with my macro fix to double check that my local installation isn't being weird. I'm expecting a runtime error on devel...


HugoGranstrom:
It did indeed fail: https://github.com/pietroppeter/nimib/actions/runs/4088239152/jobs/7049697895

pietroppeter:
this was a test, now that we have a working version I guess we can close this

PR #173, opened 2023-02-08
Change toml lib + fix macro bug on devel

HugoGranstrom:
- Fixes the orc bug on #169 
- Changes the toml parsing library to [parsetoml](https://github.com/NimParsers/parsetoml) which should make nimib run on orc finally. I had to do some workarounds though, like making my own `customToJson` proc to get the correct JSON for `std / json` to be able to parse it correctly into a `NbConfig`.

### DIFF
--- a/nimib.nimble
+++ b/nimib.nimble
@@ -13,6 +13,7 @@ requires "tempfile >= 0.1.6"
 requires "markdown >= 0.8.1"
 requires "mustache >= 0.2.1"
 requires "parsetoml >= 0.7.0"
+requires "jsony >= 1.1.5"
 
 task docsdeps, "install dependendencies required for doc building":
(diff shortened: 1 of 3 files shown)


pietroppeter:
Looks great, thanks! And loadTomlSection will be used by nimibook (see https://github.com/pietroppeter/nimibook/blob/main/src/nimibook/configs.nim, I will make the changes).

It is all here right? We can merge this and then rebase the CI Pr on the new master, correct?

HugoGranstrom:
I found one bug that the stdlib doesn't like missing fields, so I opted for using jsony instead for the deserialization. But now it should be ready to go if the CI is green. I have tried it locally with devel and the configuration is loaded without any error. So we can merge this and see if the CI PR works with it :)

pietroppeter:
good yeah, I imagined jsony would be a better fit being a loose parser. looks good, merging and rebasing CI since I am curious to see it green (then I will also add a changelog for this)

PR #176, opened 2023-02-09
updated changelog with 0.3.6 release notes + change in changelog workflow

pietroppeter:
- updated changelog with 0.3.6 release notes
- change in changelog workflow (we do not update the changelog in every PR we use automatic release notes but still update changelog after the release)
- bump version to 0.3.7 (so we are ready for next release)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,10 +5,37 @@ The changelog includes a bit of the early history of nimib, pointers to relevant
 examples of usage of nimib and heartfelt thanks to some of the fine folks that
 made this development possible.
 
-When contributing a fix, feature or example please add a new line to briefly explain the changes. It will be used as release documentation here: https://github.com/pietroppeter/nimib/releases
+For contributors: make sure to have a relevant and clear message for your PR and when merging update the first PR message with relevant notes regarding the PR (these notes will be used by maintainers during release, see below).
 
-## 0.3.x
-* _add next change here_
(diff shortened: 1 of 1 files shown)


pietroppeter:
let me know if you have notes on this @HugoGranstrom but should be rather straightforward.

pietroppeter:
added some notes also on good first issues (started defining what we mean and labeling them, there might be others) and on nimib speaking hours (created the new discussion forum category). Also updating the meta task in nimib organization on active contributors hunt: https://github.com/orgs/nimib-org/projects/1/views/1?pane=issue&itemId=19136663

HugoGranstrom:
This PR looks good 👍 :D 
Looked through the hunting contributors part quickly and it looks like reasonable things. Will sit down and take a closer look at all that has happened today (alot 🤯) so I'll get back to you then if I find anything more in particular.

HugoGranstrom:
The "bump version directly after release" is genius! :exploding_head: We will never (well...) have to forget bumping the version before a release again :O

 Regarding contributors, there seems like Hacktoberfest will be our real big push this year and until then we will prepare resources for new contributors? 
I like the idea of `all-contributors` but I find it a bit hard to read one of those because of emoji overload :sweat_smile: But it might just be something you have to be accustomed to until you understand what each emoji means. And I suppose we won't use many of the emojis in the beginning either.

Issue #117, opened 2022-07-13
NbBlock as a container


pietroppeter:
currently you cannot nest block inside one another e.g. you cannot do stuff like:

```nim
nbCode:
   nbRawOutput: "<div>"
   echo "hello"
   nbRawOutput: "</div>
```

the above is not a great use case, a better use case would be to create [html tabs](https://www.w3schools.com/howto/howto_js_tabs.asp):

```nim
nbTabs:
  tab("example 1"):
    nbText: "my example 1"
    nbCode: discard
  tab("example 2"):
    nbText: "my example 2"
    nbCode: discard
  tab("example 3"):
    nbText: "my example 3"
    nbCode: discard
```

and also nimislides has a similar use case when it does:

```nim
slide:
  slide:
    ...
  slide:
    ...
```

as a first thought there should be 3 changes to make:
- change block making it a variant type based on a `isContainer` field. If it is container you have fields `blocks: seq[NbBlock]`, `blk: NbBlock` as in `NbDoc`
- need also to change `nb` object to track who is current container, either `nb` itself or some container block (so that the newNbBlock template can be called with correct container where to add the block)
- finally we will need to change rendering, which should be straightforward

it is possible that this does not break too much the api, so it could be considered for 0.3.x

pietroppeter:
yes, that would be a possible issue (and the advice might just be: don't do that 😛), although in principle the current template `captureStdout` that we use might just work for that (it takes current stdout file pointer, creates a new file pointer when capturing then it puts it back).

HugoGranstrom:
Haha ok then 🤣 
Ohh 😮 yes that might actually work, but we would have to do something about the echos when creating a block as they will be captured by nbCode no matter how we do it. (Unless we disarm it around the echos specifically somehow without ignoring the bodies stdout)

pietroppeter:
> we would have to do something about the echos when creating a block as they will be captured by nbCode no matter how we do it.

ah right. in principle they could be not echos but something that is based on a stdout tracked initially... but honestly I do think that (at least in the beginning) we might just discourage people of nesting inside `nbCode` (or in general when captureStdout is involved).

HugoGranstrom:
[...]

Issue #74, opened 2021-11-17
nbShow for images and tables


pietroppeter:
we should have a dedicated command nbShow whose behaviour will depend on the type of object is applied.

For example:
- when applied to a ggplot, it should show the image (would it be possible/advisable to save as svg and incorporate directly in html?)
- when applied to a dataframe, it should print a nice table

this should be in some nimib/show module that should uses `when compiles(import ggplotnim)` or similar mechanism not to introduce additional dependencies.

I plan to do this after #24 (hopefully I will start soon working on that, likely starting from scratch)

HugoGranstrom:
> this should be in some nimib/show module that should uses when compiles(import ggplotnim) or similar mechanism not to introduce additional dependencies.

So it would only work for types that are officially supported by nimib? I don't have full understanding of how this works, but wouldn't something like this perhaps work?
```nim
template nbShow(t: untyped) =
	t.nbShowApi(nbDoc)

proc nbShowApi(m: MyCustomType, nb: NbDoc) =
	# Logic here
```
All types which implement `nbShowApi` are then supported. There might be aspects I'm missing of course which require tight nimib integration 😄

pietroppeter:
Still too early for me to understand how it will be for a custom type, but there will be some mechanism like the one you suggest to make this something that any custom type can implement.

I do think though that for each nbShow there might be custom parameters (e.g. max rows and max columns for a table, caption for a plot, ...).

HugoGranstrom:
> I do think though that for each nbShow there might be custom parameters (e.g. max rows and max columns for a table, caption for a plot, ...).

That's a very good point! Didn't think about that.

pietroppeter:
have been thinking about how this could be implemented, here I describe a possible implementation:

1. change partial for `nbCode` from
```nim
nb.partials["nbCode"]= """
{{>nbCodeSource}}
{{>nbCodeOutput}}
{{>nbCodeHtmlOutputs}}
"""
nb.partials["nbCodeHtmlOutputs"]= """
{{#htmls}}{{.}}{{/htmls}}
"""
```

2. we provide a `addHtml(blk: var NbBlock, html: string)` function that adds a generic html string to current block (initializes sequence `htmls` if not present.

3. a generic `nbShow[T](obj: T)` template is provided that calls a `T.toHtml` function and uses `addHtml` to add output to current block.

4. for every specific type that ones want to show, you need to implement `toHtml(o: MyType): string =`

[...]

PR #177, opened 2023-02-12
implement container blocks

HugoGranstrom:
This is my attempt at implementing container blocks (#117) thus far.

TODO:
- [ ] Refactor all relevant blocks that have bodies to support container blocks.
- [ ] Rewrite `nbCollectAllNbJs` to work with arbitrary numbers of nested levels of blocks. (recursive)
- [ ] Replace `nimibCode` with `nbCode` in docs.
- [ ] Verify that it doesn't break any current code 
- [ ] Everything else I haven't thought of...

### DIFF
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -52,6 +52,9 @@ template nbInit*(theme = themes.useDefault, backend = renders.useHtmlBackend, th
   nb.partials = initTable[string, string]()
   nb.context = newContext(searchDirs = @[]) # templateDirs and partials added during nbSave
 
+  nb.currentStdout = stdout.getFileHandle()
+  nb.originalStdout = dup(nb.currentStdout)
+
   # apply render backend (default backend can be overriden by theme)
   backend nb
 
@@ -67,6 +70,17 @@ template newNbCodeBlock*(cmd: string, body, blockImpl: untyped) =
 template newNbSlimBlock*(cmd: string, blockImpl: untyped) =
   # a slim block is a block with no body
   newNbBlock(cmd, false, nb, nb.blk, "", blockImpl)
+
+template nbUseCurrentBlockAsContainer*(body: untyped) =
+  let currentBlk = nb.blk
+  let currentBlocks = nb.blocks
+  nb.blocks = @[]
+
+  body
+
+  nb.blk = currentBlk
+  nb.blk.blocks = nb.blocks
+  nb.blocks = currentBlocks
 
 # block templates
 template nbCode*(body: untyped) =
--- a/src/nimib/blocks.nim
+++ b/src/nimib/blocks.nim
@@ -1,5 +1,5 @@
 import std / [macros, strutils, sugar]
-import types, sources
+import types, sources, capture
 
 macro toStr*(body: untyped): string =
   (body.toStrLit)
@@ -17,7 +17,8 @@ func nbNormalize*(text: string): string =
 # note that: '\c' == '\r' and '\l' == '\n'
 
 template newNbBlock*(cmd: string, readCode: static[bool], nbDoc, nbBlock, body, blockImpl: untyped) =
-  stdout.write "[nimib] ", nbDoc.blocks.len, " ", cmd, ": "
+  disableCaptureStdout:
+    stdout.write "[nimib] ", nbDoc.blocks.len, " ", cmd, ": "
   nbBlock = NbBlock(command: cmd, context: newContext(searchDirs = @[], partials = nbDoc.partials))
   when readCode:
     nbBlock.code = nbNormalize:
@@ -25,9 +26,11 @@ template newNbBlock*(cmd: string, readCode: static[bool], nbDoc, nbBlock, body, 
         toStr(body)
       else:
         getCodeAsInSource(nbDoc.source, cmd, body)
-  echo peekFirstLineOf(nbBlock.code)
+  disableCaptureStdout:
+    echo peekFirstLineOf(nbBlock.code)
   blockImpl
-  if len(nbBlock.output) > 0: echo "     -> ", peekFirstLineOf(nbBlock.output)
+  disableCaptureStdout:
+    if len(nbBlock.output) > 0: echo "     -> ", peekFirstLineOf(nbBlock.output)
   nbBlock.context["code"] = nbBlock.code
   nbBlock.context["output"] = nbBlock.output.dup(removeSuffix)
   nbDoc.blocks.add nbBlock
--- a/src/nimib/capture.nim
+++ b/src/nimib/capture.nim
@@ -5,8 +5,18 @@ import tempfile
 # Thanks, Clonk!
 
 # low level, should be Posix only but they happen to work on (my) Windows, too!
-proc dup(oldfd: FileHandle): FileHandle {.importc, header: "unistd.h".}
+proc dup*(oldfd: FileHandle): FileHandle {.importc, header: "unistd.h".}
 proc dup2(oldfd: FileHandle, newfd: FileHandle): cint {.importc, header: "unistd.h".}
+
+template disableCaptureStdout*(body: untyped) =
+  # want to write to stdout but not to our custom file.
+  let currentStdout = dup(nb.currentStdout)
+  discard dup2(nb.originalStdout, nb.currentStdout)
+  flushFile stdout
+  body
+  flushFile stdout
+
+  discard dup2(currentStdout, nb.currentStdout)
 
 template captureStdout*(ident: untyped, body: untyped) =
   ## redirect stdout to a temporary file and captures output of body in ident
--- a/src/nimib/renders.nim
+++ b/src/nimib/renders.nim
@@ -4,11 +4,19 @@ export escapeTag # where is this used? why do I export? a better solution is to 
 import highlight
 import mustachepkg/values
 
+proc render*(nb: var NbDoc, blk: var NbBlock): string
+
 proc mdOutputToHtml(doc: var NbDoc, blk: var NbBlock) =
   blk.context["outputToHtml"] = markdown(blk.output, config=initGfmConfig()).dup(removeSuffix)
 
 proc highlightCode(doc: var NbDoc, blk: var NbBlock) =
   blk.context["codeHighlighted"] = highlightNim(blk.code)
+
+proc renderContainer(doc: var NbDoc, blk: var NbBlock) =
+  var children: seq[string]
+  for child in blk.blocks.mitems: 
+    children.add doc.render(child)
+  blk.context["blocks"] = children
 
 
 proc useHtmlBackend*(doc: var NbDoc) =
@@ -52,6 +60,7 @@ proc useHtmlBackend*(doc: var NbDoc) =
   doc.renderProcs["mdOutputToHtml"] = mdOutputToHtml
   doc.renderProcs["highlightCode"] = highlightCode
   doc.renderProcs["compileNimToJs"] = compileNimToJs
+  doc.renderProcs["renderContainer"] = renderContainer
 
 proc useMdBackend*(doc: var NbDoc) =
   doc.partials["document"] = """
--- a/src/nimib/types.nim
+++ b/src/nimib/types.nim
@@ -8,6 +8,7 @@ type
     code*: string
     output*: string
     context*: Context
+    blocks*: seq[NbBlock]
   NbOptions* = object
     skipCfg*: bool
     cfgName*, srcDir*, homeDir*, filename*: string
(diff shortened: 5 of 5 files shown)


pietroppeter:
> Refactor all relevant blocks that have bodies to support container blocks.

Can't we put useCurrentBlockAsContainer directly in newNbBlock?

HugoGranstrom:
It screws up captureStdout. captureStdout must be placed outside nnContainer or else `nb.blk.output` will point to the wrong block:/

HugoGranstrom:
To expand on this, take this psuedocode:
```nim
nbContainer:
  captureStdout(nb.blk.output):
    body # new blocks are create here, thus modifying nb.blk
    # here we set the captured string:
    nb.blk.output = readFile(tempfile)
  # only here is nb.blk reset to the parent 
```
The problem stems from the fact that `captureStdout` is a template and only receives the ident `nb.blk.output`, not the value nb.blk.output had when we are calling `captureStdout`. By the time we assign to it, it has changed value and is now pointing to another block's `output` field. What's worse is that we can't save the current `nb.blk.output` in a variable at the top of `captureStdout` (before running `body`) and assign to it afterword, because `string`s are value-types in Nim. 

So if we want to use `useCurrentBlockAsContainer` directly in `newNbBlock` we would have to change `captureStdout` to take in the block instead (`NbBlock` is a `ref object`) so we can save it at the top and modify it thanks to the reference semantics. I think this is the way forward, yes we do limit `captureStdout` to not work with arbitrary strings, but I don't know why you would want to use `captureStdout` from nimib if it wasn't to display it in a block :shrug: 

The alternative is that we would have to rewrite all blocks using `captureStdout` to save the current `nb.blk` manually and pass it instead:
```nim
nbContainer:
  var blk = nb.blk
  captureStdout(blk.output):
    body
```
But this is something which is error prone if you forget about it and hard to debug.

pietroppeter:
Thanks for the explanations. It seems captureStdout plays a very special role and with these changes an even more special role (since we are introducing disabling it). I am wondering if maybe a better option would be to manage it at nbDoc level. We could have a sequence of outputs there where we add every time we catch. We then should find a way to save indices of capture for nbCode and any blocks that captures. It looks like this approach would break a lot more stuff though.

pietroppeter:
Ah yeah maybe it does not help here and it was an orthogonal thought and we can just discard it. 

No problem in changing captureStdout to take a NbBlock, we might want to keep also the string version there since it might in general be useful to people even if it is not used anymore by nimib (it is in fusion but fusion is not in a good shape).

HugoGranstrom:
Ah yes that might work, maybe we don't require the `ident` to be `untyped` but can make one overload with `NbBlock` and one with `string`. The string version could be used by the NbBlock one. This would cause minimal breakage as current code would continue to work as before. Great idea! 😄

pietroppeter:
Mmh, but what about now that nbCode is/will be a container block we have outputs that are actually blocks? Every time you capture output you add it as a block. In this way we would be able to alternate output with images and other stuff. Not sure why I keep coming to this (I do not really feel it is a very important feature), I am just trying to find a more natural api (and I give it 50-50 that I am wrong again... ;))

HugoGranstrom:
I think I see what you mean. If we do
```nim
nbCode:
  echo 1
  nbshow img
  echo2
```
it should show the "1", then the image, and last the "2"? It would be nice, but how would we tell captureStdout to split the capture everytime we add a new block (and how do we know we have added a new block)?

pietroppeter:
> how would we tell captureStdout to split the capture everytime we add a new block (and how do we know we have added a new block)?

Every time we end a capture (for example disabling it) we add a new block like we add a block when we show an image. If this is doable...

HugoGranstrom:
So `nbShow` and every block you would want to show interlaced with captured text would have to use disableCaptureStdout? What about the disableCaptureStdouts in newNbBlock, that would split any text at least 2 time 🤔 I'm not saying it doesn't work, I'm just not sure how it would work.

pietroppeter:
Basically we are not using anymore the output field of nbCode block

HugoGranstrom:
Hold on, I contradicted myself there, nbShow wouldn't have to manually use disableCapture as newNbBlock does it for us already. Then we just have the problem that we will be creating quite a few empty blocks if there isn't actually any blocks being created.

HugoGranstrom:
> Basically we are not using anymore the output field of nbCode block

So far I'm following along, it's the specifics of how to split the capture into blocks that I'm having problems with 😅

HugoGranstrom:
Let's take my example from above:
```nim
nbCode:
  echo 1
  nbshow img
  echo 2
```
`nbShow` being a block has two `disableCaptureStdout` (because it has one `stdout.write` before `blockImpl` and one after). This the list of blocks of the `nbCode` would become:
```nim
[
  NbBlock("1") # psuedo code for the block containing the captured string "1",
  # The first `disableCaptureStdout`
  NbBlock("img"),
  # The second `disableCaptureStdout`
  # nbShow didn't capture anything so this one is empty:
  NbBlock(""),
  NbBlock("2")
]
``` 

It is the stray `NbBlock("")` that I'm thinking of. How would it be created and rendered? We will create one of these `NbBlock("")` for **every block that is created in a container**. 

If we are going this route, I think we should have a dedicated template for stopping the capturing and resuming it, and not rely on `disableCaptureStdout` to handle it for us. 

I'm still not entierly convinced that this is something that we want to pursue, it is quite different from the way we are currently doing it.

pietroppeter:
Well the block that captures nothing we can probably skip it (the moment we need to create the block that captures we check if it has actually captured something).

I do agree that we might want to have a better mechanism for managing captures. I would probably go at the level of managing this at NbDoc level than as a template but no concrete plans.

Incidentally, when I am thinking of trying to write some code about this, I am always thinking I want also to try out the NbBlock using inheritance (although unsure if it creates more problems than it solves, so I will try to resist the urge...)

HugoGranstrom:
> Well the block that captures nothing we can probably skip it (the moment we need to create the block that captures we check if it has actually captured something).

And how do we check what a block has captured, if anything at all? :sweat_smile: It would require us to introduce some way of keeping track of it. There's just too many unknowns at the moment for me to get on this plan, I need concrete examples before I can try to understand :thinking: I think the current way is nice because it is simple and clear which blocks will capture what. So an alternative approach would in my mind require that as well. 

> I do agree that we might want to have a better mechanism for managing captures. I would probably go at the level of managing this at NbDoc level than as a template but no concrete plans.

Yes we need something to handle it, but I'm not sold on doing it at the NbDoc level. What advantages would it bring? Do we ever need to access the same capture more than once? My current understanding of it, is that it just complicates rendering by scattering the data in multiple places. In my ideal world, a block should itself contain all the information needed to render itself. 

> Incidentally, when I am thinking of trying to write some code about this, I am always thinking I want also to try out the NbBlock using inheritance (although unsure if it creates more problems than it solves, so I will try to resist the urge...)

Haha I've also been a bit tempted recently :rofl: But I've come to grasps with the fact that these two problems are completely orthogonal to each other so I'm leaving the object until after this is done hehe :sweat_smile: 
I would love to see some pseudocode from you though on an example (the one I have used for example) and how everything roughly would work :) I won't be able to do anything useful until I actually understand how you are thinking :rofl:

pietroppeter:
so I tried to make explicit my ideas here in this mini nimib (a minib 😛): https://gist.github.com/pietroppeter/493ff1b546f783d12c935e7fc1e25532

I did try the object hierarchy in the end, I agree it is orthogonal but it allows me to write easily actual code (not pseudocode) that does stuff. The only problem is that capturing is buggy, have not had time to fix it.

Also, your disableCapture template is fine (it would need to be adapteb), but we might also decide to skip logging when we are capturing output.

HugoGranstrom:
Wow thanks! :D And "minib" is a word we should use more often :sunglasses: Cool that you can write a barebones nimib in like 100 lines of code :exploding_head: I think I understand what you mean by doing the capturing in the NbDoc now :+1: We don't use a template to keep track of the state, but instead use fields in the doc object and discard the idea of "resuming" a capture. Every start of a capture will check if there is a current capture and create a output-block for it. This seems very clean! I assume the 
```nim
let wasCapturing = nb.isCapturingOutput
if wasCapturing:
  nbEndCapture
```
could be hidden away in a template (maybe directly in `nbCaptureOutput`?). 

A bit off-topic, but for the object hierarchy, is this kind of rendering you want to do or will we keep the partials and contexts? Or I guess each block could decide for itself how it wants to be rendered.

pietroppeter:
Yep, I think now I explained myself better, you seem to have gotten what I vaguely meant before :)

> And "minib" is a word we should use more often 😎

Yeah, I ran into the word at the end and love it already 🥰. Actually I think I will use `mb` instead of `nb` as a prefix so that we can write nimib blogposts about any minib we come up with without name clashes (it is about time I write another nblog post).

> I assume the
`let wasCapturing …` could be hidden away in a template (maybe directly in nbCaptureOutput?).

I guess it could, no preferences on that. it is actually to be able to resume capturing if you have a nbCode inside another nbCode


> A bit off-topic, but for the object hierarchy, is this kind of rendering you want to do or will we keep the partials and contexts? Or I guess each block could decide for itself how it wants to be rendered.

I do not have concrete ideas on how to render but the OOP technique should be rather flexible. Ideally we could get rid of block partials and renderProcs. It remains to be seen if we want to go down the road though, just by playing with it for minib I found some rigidity (initially my base block was empty and I had a container block inheriting from that but I was not able to make that work).

Another possible direction for a different minib is one where all blocks are of a simple type:

```nim
type
  NbBlock = ref object
    blocks: seq[NbBlock]
    data: JsonNode
```

HugoGranstrom:
> Yeah, I ran into the word at the end and love it already smiling_face_with_three_hearts. Actually I think I will use mb instead of nb as a prefix so that we can write nimib blogposts about any minib we come up with without name clashes (it is about time I write another nblog post).

That's a good idea, that would give people a good view into the basics of how nimib itself work as well. So it could work as a primer for people interested in contributing :D

> I guess it could, no preferences on that. it is actually to be able to resume capturing if you have a nbCode inside another nbCode

Reducing boilerplate for writing a block that captures is a good thing in my book, it is less error-prone as you don't have to remember to add all these steps (check if capturing, end capture, start capture, remember to end capture when done). Exactly how we the API will look like is another question though.

> Ideally we could get rid of block partials and renderProcs.

Unless we replaced these mechanics with something else, we would lose the ability to customize an existing block (using partials) and to reuse procs used by other blocks. We have an issue open in [nimiSlides](https://github.com/HugoGranstrom/nimiSlides/issues/13) to move from `nbRawHtml` to real blocks for this exact reason. The procs would probably be manageable to get working in another way, but I don't see another way of customizing how an existing block is rendered :/ It's not like you can override the `render` method of `NbCodeBlock`, that would just cause `anambigous symbol` errors if defined in a separate library. 

> Another possible direction for a different minib is one where all blocks are of a simple type:

Yes that's very reminiscent of the way we are doing it now but with different types, so this would be a good one to showcase if we keep the current way.

HugoGranstrom:
What is the take-away from these discussions? :rofl: Should we wait a bit until we proceed with container blocks and do them at the same time we switch to OO blocks? It is 2023H2 goal after all.

pietroppeter:
Well I always thought container blocks could come before the refactoring. Also still not sure about going the OO way, so I would keep the two things separate. No pressure on trying to get container blocks out now. We definitely found out they are more complex then expected (especially because of the capturing output part). We could actually implement container blocks but with limitations on output capture (you cannot/should not put nbCode inside nbCode), maybe that's a good compromise for now.

pietroppeter:
this anticipated discussion on container blocks though started because of something we wanted to do 2023H1 (NbShow #74) and there we argued container blocks could be an enabler for a better api for that.

HugoGranstrom:
> Well I always thought container blocks could come before the refactoring. Also still not sure about going the OO way, so I would keep the two things separate. No pressure on trying to get container blocks out now. We definitely found out they are more complex then expected (especially because of the capturing output part). We could actually implement container blocks but with limitations on output capture (you cannot/should not put nbCode inside nbCode), maybe that's a good compromise for now

Ok, as adding container blocks without taking capturing into account doesn't change any defined old behavior, it should be ok to add it in a minor release. What's still to be decided though is if we actually want a `nbUseCurrentBlockAsContainer` template that you have to use in all blocks supporting containers. We still need it to handle capturing correctly regardless of if we do nested capturing or not...

> this anticipated discussion on container blocks though started because of something we wanted to do 2023H1 (NbShow https://github.com/pietroppeter/nimib/issues/74) and there we argued container blocks could be an enabler for a better api for that.

Haha true, maybe I should start in that end :rofl: I'll not be available during the weekend, so I'll probably not be able to get any work done until next weekend. Maybe I will let this PR sit for a few weeks.

PR #179, opened 2023-02-22
Implement nbShow and highlight.js support for code in markdown

### LINKED ISSUE #68: support for highlighting other languages
with incoming PR #64 we will start to see more languages in nimib.

we should make again easily available the use of highlight.js in documents which contain other languages (initially nimib did nim language highlighting through highlight.js, then thanks to #15, it passed to do it statically).

this would also ease the support of nim highlighting in pure markdown pages (such as those used in nimibook, see examples there) which is not available statically (it would require patching markdown generator or postprocess the html)

implementation should be rather straightforward. among (minor) steps to be understood how to do:
- [ ] how do we avoid highlight.js to work again on nim code blocks already highlighted?
- [ ] which highlight.js set of languages we will provide? (nim is not in any of the official packages, we could go with nim + common languages but this does not include Julia, which we would want to have) see here: https://highlightjs.org/download/)

### LINKED ISSUE #74: nbShow for images and tables

HugoGranstrom:
fixes #74 fixes #68 

- `nbShow` seems to work fine, see penguins.
- highlight.js seems to affect the statically highlighted code... So we have to solve that somehow.
  - E.g in penguins the `echo df` isn't highlighted correctly.

### DIFF
--- a/docsrc/penguins.nim
+++ b/docsrc/penguins.nim
@@ -17,7 +17,10 @@ nbCode:
   let df = readCsv("data/penguins.csv")
 nbText: "let us see how it looks"
 nbCode:
-  echo df # why so wide the first column?
+  echo df
+nbText: "or even nicer using `nbShow(df)`:"
+nbShow(df.head(10))
+
 nbText: """Note that among the 7 columns of the dataframe, the first 2 and the last one have datatype string.
 The remaining 4 are numeric but they have datatype object. Why?
 
--- a/docsrc/penguins.nim
+++ b/docsrc/penguins.nim
@@ -18,7 +18,7 @@ nbCode:
 nbText: "let us see how it looks"
 nbCode:
   echo df
-nbText: "or even nicer using `nbShow(df)`:"
+nbText: "or even nicer using `nbShow(df.head(10))`:"
 nbShow(df.head(10))
 
 nbText: """Note that among the 7 columns of the dataframe, the first 2 and the last one have datatype string.
--- a/README.md
+++ b/README.md
@@ -236,15 +236,11 @@ All the options available can be seen by running any nimib file with option `nbH
 ```nim
 import osproc
 withDir nb.srcDir:
-  echo execProcess(&quot;nim r --verbosity:0 --hints:off hello --nbHelp&quot;)
+  echo execProcess(&quot;nim r --verbosity:0 --hints:off --warnings:off hello --nbHelp&quot;)
 ```
 
 
 ```
-/home/hugo/code/nim/nimib/src/nimib/sources.nim(1, 11) Warning: imported and not used: 'sugar' [UnusedImport]
-/home/hugo/code/nim/nimib/src/nimib/renders.nim(1, 12) Warning: imported and not used: 'os' [UnusedImport]
-/home/hugo/code/nim/nimib/src/nimib/renders.nim(1, 12) Warning: imported and not used: 'strformat' [UnusedImport]
-/home/hugo/code/nim/nimib/src/nimib.nim(201, 3) Warning: Use nbCodeDisplay instead; nbJsShowSource is deprecated [Deprecated]
 Nimib options:
 
   --nbHelp,     --nimibHelp                 print this help
--- a/README.md
+++ b/README.md
@@ -191,6 +191,14 @@ and [counter](https://pietroppeter.github.io/nimib/counters.html) for examples o
 In [caesar](https://pietroppeter.github.io/nimib/caesar.html) we have an example of a karax app
 that implements [Caesar cipher](https://en.wikipedia.org/wiki/Caesar_cipher).
 
+### highlighting
+Code blocks produced by `nbCode` are statically highlighted, but code in markdown code blocks are dynamically highlighted using 
+[highlightjs](https://highlightjs.org/). The dynamic highlighting can be disabled by running `nb.disableHighlightJs()`. 
+The supported languages are the ones listed as "common" [here](https://highlightjs.org/download/) plus Nim, Julia and Latex.
+
+Highlight styling classes are the same of [highlightjs](https://highlightjs.org/)
+and you can pick a different styling (`atom-one-light` is default for light mode, `androidstudio` is default for dark mode).
+
 ### latex
 
 See [numerical](https://pietroppeter.github.io/nimib/numerical.html) for an example of latex usage.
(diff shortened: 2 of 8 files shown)


HugoGranstrom:
One potential solution to the highlight.js problem is to give the user two modes:

1. Statically highlight Nim code but do not allow highlight.js to highlight Nim code (Nim code in markdown isn't highlighted).
2. Everything is highlighted using highlight.js.

HugoGranstrom:
Nevermind, found that you can mark a code block as `nohighlight` :tada:

pietroppeter:
Will look into it and let you know. As a first quick feedback I would probably keep `nim` in code class (not sure why I put hljs, was it needed for the css highlighting?)

pietroppeter:
Some more feedback:
- should we make this something that we can opt out? It would require creating a new line in head partial and a new data point highlightJs in context that can be turned off. Putting next to the css feels a bit of a hack ;)
- do we have somewhere in the documents an example where we have Nim code in a markdown context, so that we can see it in action?

pietroppeter:
(Btw not even sure if this is ready for review...)

HugoGranstrom:
> As a first quick feedback I would probably keep nim in code class (not sure why I put hljs, was it needed for the css highlighting?)

In my minimal example adding the `nim` class doesn't make a difference when using `nohighlighting`, but adding `hljs` ruins it. So I could add `nim` back. 

> do we have somewhere in the documents an example where we have Nim code in a markdown context, so that we can see it in action?

Interactivity has quite a few markdown code blocks. As it stands, the two different code blocks have different styles though :/ 

> should we make this something that we can opt out? It would require creating a new line in head partial and a new data point highlightJs in context that can be turned off. Putting next to the css feels a bit of a hack ;)

Yeah yeah yeah, we can add that when cleaning up at the end ;)

> (Btw not even sure if this is ready for review...)

It's absolutely ready for review :) And changes, and a new review after them ;)

HugoGranstrom:
Hold on `hljs` is the only thing we have in the CSS file, so `nim` is the one we don't need. Trying to add `hljs` back, seems like it doesn't break anymore locally :rofl:

pietroppeter:
> Interactivity has quite a few markdown code blocks. As it stands, the two different code blocks have different styles though 

I have looked now and it seems it comes out fine.

I would keep both nim and hljs in the class since nim is a semantic element (and hljs needed for css). And nohighlite should be enough to making it avoid doing the work twice (have not looked if there are error in console though).

I see the tests are failing but that is because we hard coded the nbCode expected output and we changed it.

pietroppeter:
nbShow looks fine!

HugoGranstrom:
The only difference I have found so far is that we are highlighting common functions in the static highlighting that highlight.js is not (above is static, below is dynamic highlighting):
image
I don't think this difference matters that much tbh.

> I would keep both nim and hljs in the class since nim is a semantic element (and hljs needed for css). And nohighlite should be enough to making it avoid doing the work twice (have not looked if there are error in console though).

Ok, will keep them then :+1:

docsrc/penguins.nim
```diff
@@ -17,7 +17,10 @@ nbCode:
   let df = readCsv("data/penguins.csv")
 nbText: "let us see how it looks"
 nbCode:
-  echo df # why so wide the first column?
+  echo df
+nbText: "or even nicer using `nbShow(df)`:"
```
pietroppeter, line 21:
```suggestion
nbText: "or even nicer using `nbShow(df.head(10))`:"
```
for consistency with what we actually run

pietroppeter (approved):
Apart from a minor suggestion, it looks good!

One last detail:
- the highlight.min.js where is it taken from? from the past? If so I think in the past I selected a number of languages that are supported, not only nim (I see python and rust in the code, for example) I think I might have put the list in the documentation in the past.
- we probably should mention somewhere this new highlight feature (available not only for nim but for other languages) somewhere in the docs

HugoGranstrom:
> the highlight.min.js where is it taken from? from the past? If so I think in the past I selected a number of languages that are supported, not only nim (I see python and rust in the code, for example) I think I might have put the list in the documentation in the past.

I took `common + Nim + Julia + Latex`, didn't really see any other language of interest on the list. If we want to add a new one in the future, it's just a matter of generating and uploading a new js file. 

> we probably should mention somewhere this new highlight feature (available not only for nim but for other languages) somewhere in the docs

Good idea :+1:  The question is where to put it, the readme probably?

pietroppeter:
> > the highlight.min.js where is it taken from? from the past? If so I think in the past I selected a number of languages that are supported, not only nim (I see python and rust in the code, for example) I think I might have put the list in the documentation in the past.
> 
> I took `common + Nim + Julia + Latex`, didn't really see any other language of interest on the list. If we want to add a new one in the future, it's just a matter of generating and uploading a new js file.

ah good selection.

> 
> > we probably should mention somewhere this new highlight feature (available not only for nim but for other languages) somewhere in the docs
> 
> Good idea 👍 The question is where to put it, the readme probably?

yes, I would go with readme, a new subsection inside "Features" about "highlighting". There is already a mention of static highlighting in the bullet list at the beginning of features, that might be reviewed. Having the list of languages (even as a list to common in highlightjs) would be useful for the user.

pietroppeter:
> Anything more you want to add?

nope, that's good, thanks for working on this. merging!

Next step would be to release with some appropriate notes, after that we add a commit to bump version and update changelog.

HugoGranstrom:
> Next step would be to release with some appropriate notes, after that we add a commit to bump version and update changelog.

I'm on it!

HugoGranstrom:
Have I forgotten something? For some reason it didn't allow me to create a discussion, something about there not existsing a title and description :shrug:

HugoGranstrom:
> It usually tells you that you need to create a tag to populate automatic release notes

I'm not sure I follow, there were no problems generating the automatic release notes or creating the tag. It was just creating the discussion forum post that was the problem.

pietroppeter:
Ah ok, not sure why then. Maybe you need special permissions?

HugoGranstrom:
Yeah perhaps, not sure exactly what permissions I have. That should hopefully be easier to manage in nimib-org hopefully.

Issue #183, opened 2023-02-26
add `nbCodeSkip` that displays highlighted code but does not compile or run it


pietroppeter:

dlesnoff:
Similar to #156 . We could implement both together.

neroist:
Can't this be closed due to #187?

pietroppeter:
Yep it can, thanks! As a GitHub trick if you uses closes #<number of issue> (or fixes or other keywords, see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) it will be closed automatically ;)

PR #187, opened 2023-03-16
Add `nbCodeSkip` and `allblocks.nim`

neroist:
Add `nbCodeSkip` from #183

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Currently most of the documentation on customization is given by the examples.
 * `nbRawHtml`: called with string content, it will add the raw content to document (html backend)
 * `nbTextWithCode`: a variant of `nbText` that also reads nim source. See example of usage
   at the end of the source in `numerical.nim` linked above.
+* `nbCodeSkip`: a variant of `nbCode` that that displays highlighted code but does not compile or run it.
 * `nbPython`:  can be used after calling `nbInitPython()` and it runs and capture output of python code;
   requires [nimpy](https://github.com/yglukhov/nimpy).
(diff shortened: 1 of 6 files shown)


neroist:
On the same fork I also added `nbCapture` (branch `nb-capture`), but I want to add a pull request for that after this one has been merged.

HugoGranstrom:
First and foremost, thank you so much for contributing to nimib! :smile: It was a pleasant surprise to see a PR from a new face. 

As for the code it looks good, the only thing missing (as per the [CONTRIBUTING.md](https://github.com/pietroppeter/nimib/blob/main/CONTRIBUTING.md)) is to
- [ ] add a basic test to [tests/tnimib.nim](https://github.com/pietroppeter/nimib/blob/main/tests/tnimib.nim) 
- [ ] mention it in the readme
- [ ] Showcase it somewhere in the documentation. Not sure exactly where though, we are open to suggestions. Maybe we can create a new document that is simply called `showcase.nim` where we drop showcases of blocks?

pietroppeter:
> Showcase it somewhere in the documentation. Not sure exactly where though, we are open to suggestions. Maybe we can create a new document that is simply called showcase.nim where we drop showcases of blocks?

Yep I think we should have something like that. I propose an `allblocks.nim` document where we have a brief description and example for all blocks. In this PR you could add just an example for this single block and we will have a separate issue/PR to populate it with the rest of blocks and think how we can structure that better

HugoGranstrom:
Awesome! If you are happy with this PR I'm ready to merge it :D

neroist:
Okay, I'm happy with it! It's okay to merge.

PR #189, opened 2023-03-17
Add `nbCapture`, closes #156

### LINKED ISSUE #156: Block that only shows the captured output of a code block
Right now we can do 
```nim
nbCode:
  echo "Hello world"
```
And it will show the code and the captured output. But what if we only want to show the captured output. Either we modify the `nbCode` partial so we optionally can hide the code like we can hide the output currently. Or we introduce a new block
```nim
nbCapture:
  echo "Hello world"
```

neroist:


neroist:
I messed up the commits... still though.

pietroppeter (approved):

HugoGranstrom (approved):

HugoGranstrom:
Wonderful work again @neroist 😁 Looks good to me.

PR #190, opened 2023-03-17
Add meta tag with generator, closes #170

### LINKED ISSUE #170: add meta generator nimib to head in all nimib generated files
we should add in head section of every nimib generated document something like:
```<meta name="generator" content="nimib">```

it is a common thing to have in SSG and also standard html metadata (see ref below)

it would also be nice to have version number but I guess more complex (in nim there is not a standard way to get version of a package).

for ref [in hugo](https://gohugo.io/functions/hugo/) they add: `<meta name="generator" content="Hugo 0.63.2">`

ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name#standard_metadata_names_defined_in_the_html_specification

### PR DESCRIPTION
**Pull request by neroist**
Also includes version in the tag.

### DIFF
--- a/src/nimib/themes.nim
+++ b/src/nimib/themes.nim
@@ -1,5 +1,5 @@
 from mustachepkg/values import castStr
-import types, gits, highlight
+import types, gits, highlight, config
 
 const document* = """
 <!DOCTYPE html>
@@ -20,7 +20,7 @@ const head* = """
   {{{favicon}}}
   <meta content="text/html; charset=utf-8" http-equiv="content-type">
   <meta content="width=device-width, initial-scale=1" name="viewport">
-  <meta content="nimib" name="generator">
+  <meta content="nimib {{{version}}}" name="generator">
   {{{stylesheet}}}
   {{{highlight}}}
   {{^disableHighlightJs}}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/nimib/themes.nim [hunk 1]
```diff
@@ -20,6 +20,7 @@ const head* = """
   {{{favicon}}}
   <meta content="text/html; charset=utf-8" http-equiv="content-type">
   <meta content="width=device-width, initial-scale=1" name="viewport">
+  <meta content="nimib {{{version}}}" name="generator">
```
pietroppeter, line 23: It is not going to matter for this but I would go with the two brackets only version `{{version}}`. The difference is that the triple bracket one (or equivalently `{{&not_escaped}}` and I actually started to prefer this since it is more evident it is different but nimib code base still mostly uses triple braces) does not escape content so that you can add valid html.

pietroppeter (approved):
Wow, thanks :) just a minor note on my side!

src/nimib/themes.nim
neroist: Okay, I changed it to use only two brackets.

HugoGranstrom (approved):

PR #192, opened 2023-03-21
Make "figure:" not show up when using nbImage with no caption

neroist:


pietroppeter:
good point! I checked to see if the image syntax with empty square brackets is valid and for github it is so looks good to me. btw I just realized that the stuff in empty square brackets should actually be alt text instead of caption but I guess for the moment it is not worth to adjust that.

pietroppeter (approved):

HugoGranstrom (approved):

PR #195, opened 2023-03-21
Add seperate `alt` param to  `nbImage`, close #193

### LINKED ISSUE #193: add alt text for nbImage
currently the nbImage block has only two parameters: `url` and `caption` and `caption` is used for alt text both in html and markdown backend. We should add an alt text parameter to be used correctly since caption and alt text in general should be different. We could still keep the old behaviour of using caption when alt text is missing. raised when reviewing #192

neroist:
If `alt` is empty the caption is used instead. Also added a new test in `trenders.nim`.

### DIFF
--- a/src/nimib.nim
+++ b/src/nimib.nim
@@ -101,13 +101,14 @@ template nbTextWithCode*(body: untyped) =
   newNbCodeBlock("nbText", body):
     nb.blk.output = body
 
-template nbImage*(url: string, caption = "") =
+template nbImage*(url, caption = "", alt: string = "") =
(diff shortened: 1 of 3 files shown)


HugoGranstrom (approved):
Looks good to me, thanks again 😄

May I be frank and ask you how you ended up being the nimib contributor of the month? 😉 Have you been lurking around for a while or was it the recent forum announcement that brought you here?

pietroppeter:
Thanks again! 😄

HugoGranstrom:
I see, looks nice 👍😄

PR #196, opened 2023-03-26
Fix `getNimibVersion()`

neroist:
Forgot that nimble doesn't include the `src` dir, so you'd have to only call `parentDir` twice.

### DIFF
--- a/src/nimib/config.nim
+++ b/src/nimib/config.nim
@@ -1,11 +1,12 @@
 import types, parsetoml, jsony, std / [json, os, osproc, math, sequtils]
 
 proc getNimibVersion*(): string = 
-  let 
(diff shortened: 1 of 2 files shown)


HugoGranstrom:
Would it be possible to write a test for this? Just checking that it is a SemVer version or even just that it contains one or two `.`.

neroist:
Sure!

pietroppeter (approved):

pietroppeter:
sorry closed by mistake. Thanks for this! I guess you noticed when trying to deploy https://neroist.github.io/nimitheme/ ? I see it is working fine the meta tag with nimib version now, so I would assume it is already making use of the fix.

PR #199, opened 2023-03-27
bump 0.3.10 and update changelog with 0.3.9 release notes

pietroppeter:


pietroppeter:
to be honest not sure if it is worth the effort of duplicating the content of release notes in changelog.md anymore. Should we discontinue changelog.md (we will also decide to go bak to it if we don't like the decision). Thoughts @HugoGranstrom? I could add a note in changelog already in this PR

HugoGranstrom:
I actually think changelog.md is more important than the release page. It's easier to browse and search for when a feature was introduced/changed. Is it really that much effort? Copy-paste and commit? 😅

pietroppeter:
nope, it is really low effort. let's keep it like that then!

PR #202, opened 2023-03-28
Refactor tests and remove hints to better visualize success/failure

dlesnoff:
The for loop on the list of file names is more readable.
I added hints:off compilation option, to remove superfluous information for the tests. (Similarly to [bigints](https://github.com/nim-lang/bigints/blob/master/bigints.nimble))

To be consistent in syntax, I can do the same for documentation building.

### DIFF
--- a/docsrc/numerical.nim
+++ b/docsrc/numerical.nim
@@ -18,7 +18,7 @@ proc introText(defaultStyle = false): string =
 > This nimib example document shows how to:
 >  - apply (or not) a custom style ([latex.css](https://latex.now.sh/)){otherStyle}
 >  - using latex rendering of equations (thanks to [katex](https://katex.org/))
->  - having as output an html table (using standard markdown converter to table) 
+>  - having as output an html table (using standard markdown converter to table)
 >
 > The document itself shows how to use [numericalnim](https://github.com/hugogranstrom/numericalnim)
 > to integrate an ODE.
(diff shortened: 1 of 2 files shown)


pietroppeter (approved):
Looks good to me!

HugoGranstrom (approved):
Nice, a well needed refactor that wet should probably do for the docs as well as you said.

pietroppeter:
> There is an issue with numericalnim which gets in the way when building numerical.nim so I removed it from the list.
> I mentioned some of the specific dependencies I had to install (due to imports in the documentation) in a comment.
> It would be nice if this task activates extra requirements so that nimble can do the proper installations automatically.

Not that there is already a task docdeps that specifies which version of the libraries are needed to build the docs. As such there is no need for an extra comment to mention dependencies, nor I think there is a need to remove numerical from docs.

dlesnoff:
I reverted last commit since I did not see that extra task.
I removed numerical to build other docs, since it stops at the very first build failure.

HugoGranstrom:
The problem with numerical.nim is that it uses an old version of numericalnim. If you have a more recent version installed it will use that instead and it will fail to compile. I think it is the `NumContext[float]` that should be `NumContext[float, float]` for it to work with the latest numericalnim.

dlesnoff:
I reverted the numericalnim changes.

pietroppeter:
Thanks again!

dlesnoff:
You are welcome!

PR #203, opened 2023-03-28
Complete Allblocks

dlesnoff:
Fix #194.
The `nbFile` and `nbShow` blocks are missing in the documentation but could be added after the PR.
I am not sure of their usage and was not very creative for examples today.

I added a TOC to list all blocks at the top and give quick access to their documentation.

The whale image (chosen for some kind of consistency with the icon) is licensed under [the Unsplash license](https://unsplash.com/license).

The documentation now requires Nimpy to work. which makes nimpy a *soft* dependency.

I removed the blocks description from the README.md.
The documentation link won't work until the docs are added to pietroppeter's blog.

Other partials (blocks too?) in `renders.nim` to add in `allblocks.nim`:
- [x] nbText
- [x] nbCode
- [x] nbFile
- [ ] nbJsFromCode
- [x] nimibCode

PR #208, opened 2023-06-10
Allblocks: add nbShow, nbCodeWithText, nbCodeInBlock

### LINKED ISSUE #194: finalize allblocks.nim
a `allbocks.nim` file was introduced in #187 to document a new code block.
it should be:
- [ ] completed documenting all existing blocks
   - [ ] nbCodeInBlock
   - [ ] nbTextWithCode
- [ ] built when building docs (add a line in nimble file, see https://github.com/pietroppeter/nimib/blob/ce52a7613a4ac6ca5f80b5817abd018d919f745e/nimib.nimble#L33)
- [ ] linked in readme/index
- [ ] also a note should likely be added in CONTRIBUTING.md about its existence

dlesnoff:
I improved a little bit the documentation of blocks by adding the documentation of the three last blocks, as discussed during the Nimib discussions meeting. Fix #194


docsrc/allblocks.nim
```diff
@@ -66,6 +65,25 @@ nimibCode:
   nbCodeSkip:
     exit() # even this won't execute!
 
+nbCodeBlock: "nbCodeInBlock"
+nbText: """
+Sometimes, you want to show similar code snippets with the same variable names and you need to declare twice the same variables.
+You can do so with `nbCodeInBlock` which nests your snippets inside invisible blocks.
```
HugoGranstrom, line 71:
```suggestion
You can do so with `nbCodeInBlock` which nests your snippet inside a block.
```
Good explanation, and I like the comments in the code below :D. I would just rephrase this sentence to not use "invisible".

docsrc/allblocks.nim
```diff
@@ -104,6 +127,21 @@ Here is the source code and the centered title:
 nimibCode:
   nbRawHtml: """<h4 style="text-align: center">Centered title</h4>"""
 
+nbCodeBlock: "nbShow"
+nbText: """
+Some Nim objects have been designed with an HTML renderer. If the object has an associated `toHtml()` procedure, it can be rendered with the nbShow plot.
```
HugoGranstrom, line 132:
```suggestion
Some Nim objects have been designed with an HTML renderer. Nimib allows to pretty-print objects by rendering them to HTML. If the object has an associated `toHtml()` procedure, it can be rendered with `nbShow`.
```
While the first sentence is technically true, it is not the point of the `nbShow` block. This is a suggestion on how to rephrase it, but feel free to modify it :D.

docsrc/allblocks.nim
```diff
@@ -75,6 +93,11 @@ nimibCode:
   nbCapture:
     echo "Captured!"
 
+nbCodeBlock: "nbTextWithCode"
+nbText: """
+`nbText` only stores the string it is given, but it doesn't store the code passed to `nbText`. For example, `nbText: fmt"{1+1}"` only stores the string `"2"` but not the code `fmt"{1+1}"` that produced that string. `nbTextWithCode` works like `nbText` but it also stores the code in the created block. It can be accessed with `nb.blk.code` right after the `nbTextWithCode` call. See the end of `numerical.nim` for an example.
```
HugoGranstrom, line 98: Once again, good explanation of why the block exists and differs from `nbText`. It would be nice if the `numerical.nim` could be a link to that page :D.

HugoGranstrom (changes requested):
Nice work, thank you! :smile: I think you have good explanations and examples of the code blocks. I have only a few rephrasing suggestions (they are only suggestions, feel free to change them further) and it would be nice with a link at one spot.

docsrc/allblocks.nim
dlesnoff: This was your explanation: https://github.com/pietroppeter/nimib/pull/203#discussion_r1153644530
I do not get what would be the link in that case.

docsrc/allblocks.nim
dlesnoff: Dropping entirely the first sentence seems better.
```suggestion
Nimib allows to pretty-print objects by rendering them to HTML. If the object has an associated `toHtml()` procedure, it can be rendered with `nbShow`.
```

docsrc/allblocks.nim
HugoGranstrom: Ok then I was lazy 😛 the link should be to https://pietroppeter.github.io/nimib/numerical.html but it's better if we use a relative link.

docsrc/allblocks.nim
HugoGranstrom: I think the first sentence is nice, it gives a brief description of the block instead of jumping right in.

HugoGranstrom: Ah yes sorry, I forgot to remove your original first sentence 😅 So yes, you are correct, it can be removed 👍

docsrc/allblocks.nim
dlesnoff: What about: https://github.com/pietroppeter/nimib/blob/main/docsrc/numerical.nim ?
```suggestion
`nbText` only stores the string it is given, but it doesn't store the code passed to `nbText`. For example, `nbText: fmt"{1+1}"` only stores the string `"2"` but not the code `fmt"{1+1}"` that produced that string. `nbTextWithCode` works like `nbText` but it also stores the code in the created block. It can be accessed with `nb.blk.code` right after the `nbTextWithCode` call. See the end of [numerical.nim](https://github.com/pietroppeter/nimib/blob/main/docsrc/numerical.nim) for an example.
```

HugoGranstrom: I think it makes more sense to link to the actual webpage, as it contains both what's rendered and the code (under the `View Source` button). Change the link name to just `numerical` if that makes more sense.

dlesnoff: I changed the link to the one you proposed. I fear that we can not make a relative link.

HugoGranstrom: > I fear that we can not make a relative link.

Huh, why? 🤔

HugoGranstrom: Doesn't linking to `./numerical.html` work?

pietroppeter (approved):
Thanks, looks great!

docsrc/allblocks.nim
dlesnoff: I was expecting the link to be relative to the script directory rather than the output directory. I changed all three links to relative ones.

dlesnoff:
It works great with relative link: https://db62d04b30bae0fe4498--nimib.netlify.app/allblocks.html
I guess that I should put a relative link to allblocks.html in the README.md too.

HugoGranstrom (approved):
Awesome! The links seem to work 😁 And yes, all you have to do now is add a link in `index.nim` (see how the other links are done there) and regenerate the README using `nimble readme`

docsrc/index.nim
```diff
@@ -132,7 +132,7 @@ Currently most of the documentation on customization is given by the examples.
 
 ### other blocks
 
-You can find a complete description of the code blocks along with examples in [allblocks.nim](https://pietroppeter.github.io/nimib/allblocks.html)
+You can find a complete description of the code blocks along with examples in [allblocks]({docs}/allblocks.html.
```
HugoGranstrom, line 135: You accidentally deleted the ending `)` for this link.

dlesnoff:
```suggestion
You can find a complete description of the code blocks along with examples in [allblocks]({docs}/allblocks.html).
```

HugoGranstrom:
Looks good, regenerate the README with this fixed link and I will merge it afterward :+1:

Issue #72, opened 2021-11-16
remove dependency tempfile (when nim >= 1.6)


pietroppeter:
Nim now has a tempfile module so dependency to tempfile library could be removed (when nim version >= 1.6): https://nim-lang.github.io/Nim/tempfiles.html

not really a priority, but I ran into this and I am noting it down.

pietroppeter:
fixed by #211

PR #211, opened 2023-08-09
Fix std/random bug (#204)

### LINKED ISSUE #204: import of nimib breaks expected behaviour of std / random
expected behaviour of [std / random](https://nim-lang.org/docs/random.html) is that the following should output always the same result, which it does not unless I remove nimib import:
```nim
import std / random
import nimib # if I comment this lines results are reproducible
echo rand(100)
```
this is a side effect of the dependency on tempfile, see: https://github.com/OpenSystemsLab/tempfile.nim/issues/13

this is particular bad in nimib since tempfile is called every time a new nbCode is created (a temporary file is created to save the output) and this seems to trigger a new call to randomize. This means that a single `randomize(42)` call after importing nimib is not enough to guarantee reproducible results, the user must call randomize with a fixed seed inside every code block that calls random functions.

```nim
import nimib
import std / random
randomize(42) # putting a randomize here does not guarantee reproducible results
nbInit
nbCode:
  randomize(42) # putting randomize here does!
  echo rand(100)
echo nb.blk.output
```

An alternative to tempfile are some utilities in std lib but they do not yet support the same functionalities of tempfile: https://nim-lang.org/docs/tempfiles.html

sent44:
- Remove OpenSystemsLab/tempfile.nim and replace with build-in `std/tempfiles`
Without `randomize()` that can be found in that package, the random result in user code will be consistent
- Remove low level stuff and replace with build-in `fusion/ioutils`
As it might break in newer Windows update
- Add ability to name the temp file
`reuse the same file every time` I don't know if this statement mean that you can give name the temp file and if it is already exist then open or temp file name is generate and you can use old file in some way, so I implement the first one. Should not break exist code (Didn't test against the main package)

This should not need to modify docs as it is internal stuff hide be hide `nbCode`
fixed #204

Test
```nim
import capture

var s: string

var msg = "hello"
echo msg & "1"
captureStdout(s):
    echo msg & "2"
    msg = "ciao"
echo msg & "3"

assert s == "hello2\n"

import random

echo rand(100) # result: 3
captureStdout(s, "my_random_tmp"):
    echo rand(100) # result: 89
    discard
# echo rand(100)
echo "s=", s

captureStdout(s, "my_random_tmp"):
    echo rand(100) # result: 18
echo "s=", s

captureStdout(s):
    echo "echo"
echo "s=", s
```
Expect result
```
hello1
ciao3
3
s= (lots of 89 and 18 depend on how many time you run)

s= (lots of 89 and 18 depend on how many time you run)

s=echo
```

### DIFF
--- a/src/nimib/capture.nim
+++ b/src/nimib/capture.nim
@@ -3,17 +3,14 @@ import fusion/ioutils
 import std/os
 import std/tempfiles
 
-template captureStdout*(ident: untyped, body: untyped) =
-    captureStdout(ident, "", body)
-
 template captureStdout*(ident: untyped, filename: string, body: untyped) =
     ## redirect stdout to a temporary file and captures output of body in ident
     # Duplicate stdout
     let stdoutFileno: FileHandle = stdout.getFileHandle()
     let stdoutDupFd: FileHandle = stdoutFileno.duplicate()
     
-    # Create a new temporary file or attemp to open it
-    let (tmpFile, tmpFileName {.used.}) = when filename == "":
+    # Create a new temporary file or attempt to open it
+    let (tmpFile, tmpFileName {.used.}) = if filename == "":
             createTempFile("tmp", "")
         else:
             let tmpName = getTempDir() / filename
@@ -29,7 +26,7 @@ template captureStdout*(ident: untyped, filename: string, body: untyped) =
     # Flush stdout and tmpFile, read tmpFile from start to ident and then close tmpFile
     stdout.flushFile()
     tmpFile.flushFile()
-    when filename == "":
+    if filename == "":
         tmpFile.setFilePos(0)
     else:
         discard tmpFile.reopen(tmpFileName)
@@ -38,3 +35,6 @@ template captureStdout*(ident: untyped, filename: string, body: untyped) =
     
     # Restore stdout
     stdoutDupFd.duplicateTo(stdoutFileno)
+
+template captureStdout*(ident: untyped, body: untyped) =
+    captureStdout(ident, "", body)
--- a/src/nimib/capture.nim
+++ b/src/nimib/capture.nim
@@ -1,20 +1,15 @@
 ## This submodule implements only one template to temporarily redirect and capture stdout during execution of a chunk of code.
 import fusion/ioutils
-import std/os
 import std/tempfiles
 
-template captureStdout*(ident: untyped, filename: string, body: untyped) =
+template captureStdout*(ident: untyped, body: untyped) =
     ## redirect stdout to a temporary file and captures output of body in ident
     # Duplicate stdout
     let stdoutFileno: FileHandle = stdout.getFileHandle()
     let stdoutDupFd: FileHandle = stdoutFileno.duplicate()
     
     # Create a new temporary file or attempt to open it
-    let (tmpFile, tmpFileName {.used.}) = if filename == "":
-            createTempFile("tmp", "")
-        else:
-            let tmpName = getTempDir() / filename
-            (tmpName.open(fmAppend), tmpName)
+    let (tmpFile, _) = createTempFile("tmp", "")
     let tmpFileFd: FileHandle = tmpFile.getFileHandle() 
     
     # writing to stdoutFileno now writes to tmpFile
@@ -26,15 +21,9 @@ template captureStdout*(ident: untyped, filename: string, body: untyped) =
     # Flush stdout and tmpFile, read tmpFile from start to ident and then close tmpFile
     stdout.flushFile()
     tmpFile.flushFile()
-    if filename == "":
-        tmpFile.setFilePos(0)
-    else:
-        discard tmpFile.reopen(tmpFileName)
+    tmpFile.setFilePos(0)
     ident = tmpFile.readAll()
     tmpFile.close()
     
     # Restore stdout
     stdoutDupFd.duplicateTo(stdoutFileno)
-
-template captureStdout*(ident: untyped, body: untyped) =
-    captureStdout(ident, "", body)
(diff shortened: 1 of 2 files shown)


sent44:
I am going to investigate the error for the time begin
Temporary close

src/nimib/capture.nim
```diff
@@ -1,30 +1,40 @@
...
-  ## redirect stdout to a temporary file and captures output of body in ident
-  let
-    # Duplicate stdout
-    stdoutFileno = stdout.getFileHandle()
-    stdoutDupfd = dup(stdoutFileno)
-    # Create a new temporary file
-    (tmpFile, tmpFilename) = mkstemp(mode=fmWrite)
-    tmpFileFd: FileHandle = tmpFile.getFileHandle()  
-  discard dup2(tmpFileFd, stdoutFileno)  # writing to stdoutFileno now writes to tmpFile
+    captureStdout(ident, "", body)
```
HugoGranstrom, line 7: This is a template so it might not matter, but I would like you to define this overload of `captureStdout` without `filename` below the other one. I don't know if the CI error is due to this or not.

HugoGranstrom (changes requested):
First off, congratulations and thanks for opening your first PR to nimib :partying_face: Your contribution is very appreciated! 

I have requested a few changes (one which maybe could solve the CI error). In addition to that you will also have to add `fusion` as a dependency in the `nimib.nimble` file. 

> I am going to investigate the error for the time begin
> Temporary close

No need to close this IMO, there's no rush, leave it open and fix it when you have time :D

src/nimib/capture.nim
sent44: It is possible that the complier is somehow unable to reorder the code when use with Nimib

HugoGranstrom: How do you mean? I don't think the compiler does any reordering by default.

HugoGranstrom:
> > In addition to that you will also have to add fusion as a dependency in the nimib.nimble file.
> 
> Isn't fusion already part of semi standard library that already come compiler? Like experimental features
> So I should `OpenSystemsLab/tempfile.nim` dependency from `nimib.nimble` instead
> 
> Edit: Oh https://forum.nim-lang.org/t/7608

Yeah, `fusion` was planned to function that way but now it's mostly abandoned. It does have a few useful libraries though. And yes, you should remove the old `OpenSystemsLab/tempfile.nim` dependency from the nimble file.

src/nimib/capture.nim
sent44: I also don't really know since template should not need to reorder for it to work.

HugoGranstrom: Indeed 👍 I guess it did run without problem for you locally? It could perhaps be something with the other errors that's causing it to give a bad error. So I would reorder templates, add the dependency and commit it and see what the CI says about it

sent44: Yes, it is working locally and official playground

HugoGranstrom:
All the tests are passing now, so we have solved part of the issue :partying_face: Now we have a new error though, `redefinition of decode` :/

sent44:
> Now we have a new error though, redefinition of decode :/

I have no idea ._., might need to check with `expandMacro` to see why somehow it is redefine
I am going to sleep now

HugoGranstrom:
I have managed to reduce it to a minimal example, and it's a compiler problem

```nim
template foo(x: string, body: untyped) =
  echo x
  body

template foo(body) =
  foo("Goodbye", body)
  
foo("Hello"): # works
  proc f() = discard
  echo "World"
  
foo: # doesn't work
  proc f() = discard
  echo "World"
```

The problem is that we have an overload of `captureStdout`, and the compiler needs to check which one it is that we want to call. So it must type-check all the arguments (even if one of them is `untyped`). And my guess is that in that type-checking it mistakenly saves the proc `decode` and when we actually define it later it will falsely complain about it already being defined. 

I will open an issue on Nim's issue tracker, but the workaround for us would be to don't have two overloads and instead just keep one of them (with or without filename).

HugoGranstrom:
It seems like this has been discussed here: https://github.com/nim-lang/RFCs/issues/402 And it basically won't be fixed... So we have to do a workaround, and my vote is on removing the `filename` argument as I simply don't see any use for it. And if we want to reuse a file in the future we will have to refactor this code either way, so we'll deal with that when it comes instead. Your opinion?

HugoGranstrom:
Also, you will need to merge our `main` branch into your `main-1` branch as I just fixed a piece of code in the docs that didn't compile.

pietroppeter:
Great work @sent44, very well done and thanks! And excellent analysis by @HugoGranstrom of this weird issue. I would agree we can skip the version with filename.

HugoGranstrom:
Looks good to me and the CI now. Thanks again for contributing to nimib 😁 I'll merge this in a moment

pietroppeter:
note that this also closes the older issue of #72. And reading that makes me realise that this fix "breaks" nimib for nim < 1.6. Which is fine, I think. An option would have been to keep older code under `when (NimMajor, NimMinor) < (1, 6)` but it is probably not worth it at the moment (and we could do it easily later in case there is a need). What is important we do not miss is that in the changelog for next release (3.10?) we mention this breaking change. And also I think this is the only recent change that would warrant a release, right? (the other commits if I am not wrong are improvements/changes for documentation, not in the library).

HugoGranstrom:
Oh, had totally missed that :O But I agree, 1.4.x is old by now. I think we should keep 1.6.x working as long as possible though as 2.0 brings some breaking changes that could break people's old articles using old versions of libraries. 

> What is important we do not miss is that in the changelog for next release (3.10?) we mention this breaking change. And also I think this is the only recent change that would warrant a release, right? 

Correct, next release is 3.10. These two commits are not included in the latest release either:[1](https://github.com/pietroppeter/nimib/commit/472e5d8d815c3a19e64e25b7687197a2d4ff7206) [2](https://github.com/pietroppeter/nimib/commit/dedeffd181e1791a7d58594d59b31335f28a8ae9)

PR #217, opened 2023-08-19
Fix compile error when using custom nimble env (#216)

### LINKED ISSUE #216: Compile error when using custom Nimble environment variable
https://github.com/pietroppeter/nimib/blob/dc890600a2274aaec8da71f5e16be084a7f15614/src/nimib/config.nim#L9-L11
When using `nimble dump --json` with custom Nimble environment variable, it will also output `Info: Using the environment variable: ...` at the front of json.
So when using that result string with `parseJson()`, it will error.
image

sent44:
Fix #216

Tell me if you don't want regex

### DIFF
--- a/src/nimib/config.nim
+++ b/src/nimib/config.nim
@@ -1,4 +1,4 @@
-import types, parsetoml, jsony, std / [json, os, osproc, math, sequtils]
+import types, parsetoml, jsony, std / [json, os, osproc, math, sequtils, re]
 
 proc getNimibVersion*(): string = 
   var dir = currentSourcePath().parentDir().parentDir()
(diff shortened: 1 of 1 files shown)


src/nimib/config.nim
```diff
@@ -1,14 +1,19 @@
...
+import types, parsetoml, jsony, std / [json, os, osproc, math, sequtils, re]
 
 proc getNimibVersion*(): string = 
   var dir = currentSourcePath().parentDir().parentDir()
 
   if dir.splitPath().tail == "src":
     dir = dir.parentDir()
 
-  let dumpedJson = execProcess("nimble dump --json", dir) 
+  let dumpedJson = execProcess("nimble dump", dir)
```
pietroppeter, line 9:
```suggestion
  let dumpedJson = execProcess("nimble dump --json --silent", dir) 
```

using just this change and removing all other changes should be a better fix

pietroppeter (changes requested):

pietroppeter:
hi, thanks for working on this!

As mentioned here https://github.com/pietroppeter/nimib/issues/216#issuecomment-1784271686 a better fix is just to use the `--silent` option. I added a suggestion for the line to be changed and the rest of changes should be discarded

sent44:
Fixed by #230 
I was busy at that time and then forgot about this.

PR #220, opened 2023-11-05
Add: function to add line numbers to block. Not called yet.

### LINKED ISSUE #101: possibility to add line numbers to code blocks
we should have the ability to add line numbers to code blocks, both those generate by `nbCode` and `nbFile`. Ideally it should be something customizable by single block and one could set a global option. defaults would probably be to have it disabled in code blocks produce by `nbCode` and maybe enabled for blocks produce by `nbFile`.

dlesnoff:
This PR is not finished yet. Attempt to fix #101.
to do:
- [ ] move tests to correct files
- [ ] addLineNumbersToHighlightedCode is not yet called
- [ ] Separate numbers from code in the generated HTML so that code can be copy-pasted without the numbers.

How can I call the newly added procedure `addLineNumbersToHighlightedCode`?
`AddLineNumbers` does nothing for the moment.

### DIFF
--- a/src/nimib/renders.nim
+++ b/src/nimib/renders.nim
@@ -11,6 +11,22 @@ proc highlightCode(doc: var NbDoc, blk: var NbBlock) =
   blk.context["codeHighlighted"] = highlightNim(blk.code)
 
 
+# assert addLineNumbersToHighlightedCode("""
+# func</span> decode(secret: openArray[<span class="hljs-built_in">int</span>]): <span class="hljs-built_in">string</span> =
+#   <span class="hljs-comment">## classified by NSA as &lt;strong&gt;TOP SECRET&lt;/strong&gt;</span>
+#   <span class="hljs-keyword">for</span> c <span class="hljs-keyword">in</span> secret:
+#     <span class="hljs-literal">result</span>.add <span class="hljs-built_in">char</span>(c)
+# """) == """
+# <span class="hljs-comment">1</span> func</span> decode(secret: openArray[<span class="hljs-built_in">int</span>]): <span class="hljs-built_in">string</span> =
+# <span class="hljs-comment">2</span>   <span class="hljs-comment">## classified by NSA as &lt;strong&gt;TOP SECRET&lt;/strong&gt;</span>
+# <span class="hljs-comment">3</span>   <span class="hljs-keyword">for</span> c <span class="hljs-keyword">in</span> secret:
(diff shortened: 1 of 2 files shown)


src/nimib/renders.nim
```diff
@@ -1,4 +1,4 @@
-import std / [strutils, tables, sugar, os, strformat, sequtils]
+import std / [strutils, tables, sugar, sequtils]
```
dlesnoff, line 1: These are unused imports.

docsrc/allblocks.nim
```diff
@@ -17,7 +17,7 @@ template nbCodeBlock(name:string) =
   nbToc.output.add "1. <a href=\"#" & anchorName & "\">" & name & "</a>\n"
 
 nbText: """
-> This nimib document provides a brief description and example for 
+> This nimib document provides a brief description and example for
```
dlesnoff, line 20: I'm sorry, the diff contains a lot of line changes because of some automatic tool in my Vim editor to remove whitespaces at the end of lines.

src/nimib/renders.nim
```diff
@@ -10,6 +10,23 @@ proc mdOutputToHtml(doc: var NbDoc, blk: var NbBlock) =
...
+  for i in 1..nlines:
+    ## We have line numbers in the same line than code
+    result.add("""<span class="hljs-comment"> """ & $i & "</span> " & code.splitlines()[i-1] & "\n")
+    ## but we can not copy-paste code without the numbers
+  # The following separates line numbers from code, but display is wrong
+  # result = """ <div>
+  #                 <code>""" & lineNumbers.join("") & """</code>
+  #                 <code>""" & code & """</code>
+  #              </div>
+  #          """
```
dlesnoff, line 25: I think we should wrap the code in a table or a div container so that we can copy-paste code without the numbers.

neroist:
Why not use something like <https://github.com/wcoder/highlightjs-line-numbers.js>? Though I'm not sure about the impact on nimitheme...

Using it with this stylesheet works for me:
```css
td.hljs-ln-numbers {
  text-align: center;
  color: #ccc;
  vertical-align: top;
  padding-right: 5px;

  -webkit-touch-callout: none;
  -webkit-user-select: none;
  -khtml-user-select: none;
  -moz-user-select: none;
  -ms-user-select: none;
  user-select: none;
}

td.hljs-ln-code {
  padding-left: 1em;
}

/* Fix for water.css*/
table.hljs-ln {
  width: initial;
  margin-bottom: 0;
}

table.hljs-ln tbody tr:nth-child(2n) {
  background-color: unset;
}
```

PR #224, opened 2023-12-26
Add language field for full language highlighting on html backend

beef331:
This simply adds a method for having highlight work on user defined languages, the same process likely should be done on the markdown side wherever possible. No clue the scale of the hackiness.


beef331:
Talking to @HugoGranstrom this seems like a very awful way of doing it.

Issue #168, opened 2023-01-29
refactor NbBlock to use OOP (hierarchy of objects / inheritance)


pietroppeter:
- in the past we were not fan of OOP but that was probably not a good take
- indeed we now think that for NbBlock it could make a lot of sense to have a base object and a hierarchy of objects that inherit from it
- still need to make some experience in using OOP stuff (maybe in a nimin not related lib)

HugoGranstrom:
One problem we discussed in a past Speaking Hour is the fact that we must be able to serialize all our objects to JSON. The problem being that we don't have static knowledge of the exact type of each block. If we use the dynamic dispatch functionality of Nim, we can create a `dumpHook` for the base type and then call a `method` that returns the string of the specific type:
```
import jsony

type
  Nimib = ref object of RootObj
    a, b, c: int
  NimibChild = ref object of Nimib
    d, e: float


method dump(n: Nimib): string =
  n[].toJson()

method dump(n: NimibChild): string =
  n[].toJson()

method dumpHook*(s: var string, v: Nimib) =
  s.add v.dump()

let n1: Nimib = Nimib(a: 1, b: 2, c: 3)
let n2: Nimib = NimibChild(d: 3.14, e: 4.56)

echo n1.toJson()
echo n2.toJson()
```

The next problem then becomes how we parse the JSON again without losing the type information...

HugoGranstrom:
And here the template version is:
```nim
import jsony, tables

type
  Nimib = ref object of RootObj
    a, b, c: int
    typename: string
  NimibChild = ref object of Nimib
    d, e: float

var parseDefs: Table[string, proc (s: string, i: var int): Nimib]

template registerBlock(typename: untyped) =
  parseDefs[$typename] =
    proc (s: string, i: var int): Nimib =
      var v: typename
      new v
      parseHook(s, i, v[])
      result = v

  method dump(n: typename): string =
    n[].toJson()

# neccecary to avoid compile error
method dump(n: Nimib): string =
    n[].toJson()

proc dumpHook*(s: var string, v: Nimib) =
  s.add v.dump()

proc parseHook*(s: string, i: var int, v: var Nimib) =
  # First parse the typename
  var n: Nimib = Nimib()
  let current_i = i
  parseHook(s, i, n[])
  # Reset i
  i = current_i
  # Parse the correct type
  let typename = n.typename
  v = parseDefs[typename](s, i)

registerBlock(Nimib)
registerBlock(NimibChild)

[...]

Issue #209, opened 2023-06-10
Idea: Nimib plugin system (both dynlib & ordinary import)


HugoGranstrom:
This is an idea inspired by @pietroppeter's work on [nblog](https://github.com/pietroppeter/nblog/pull/16) using JSON as an intermediate form for a static site generator and [Ark](https://github.com/dmulholl/ark) (former Ivy). To summarize, the idea is to generate the `NbDoc` as usual, but instead of rendering it at the end, it is converted and save to a JSON file. Then a separate program reads these JSON files and renders them. 

This got me thinking, how does the second program have access to all the partials, templates and renderProcs needed to render all of these files? One answer to this question is that we start to structure nimib themes and libraries as plugins. More specifically, we could say that each plugin must export a `initNimibPlugin` proc that accepts a `var NbDoc`. In it, the plugin would populate the contexts and renderProcs. This is basically how themes work in nimib today:
```nim
nbInit(theme = revealTheme)
```
Now instead, imagine that we could supply multiple of these procs, one for each plugin/theme:
```nim
nbInit(plugins = @[revealTheme, plugin1Init, plugin2Init])
```
The sweet part about this is that these `initNimibPlugin` procs could come from anywhere! They could come from a module we have imported or (and this is the important bit) it could come from a shared library. In other words; this would allow the partials, templates and renderProcs to be accessed by the rendering executable by loading them using for example `std/dynlib`. 

This might not be enough for implementing a static site generator in nimib, but it is a step towards it. And it is not that big of a change relative to what we currently have, it's mainly just that we start to recommend a certain structure of nimib libraries. 

You have probably thought about a lot of these aspects already, but I haven't found a good issue to discuss it, so I'm creating this one. Is there anything I'm getting wrong or that you have different opinions on?

HugoGranstrom:
I haven't tried playing around with shared libraries myself yet, but the way I understand it is that we would be able to load the `initNimibPlugin` proc from a shared library `plugin.so` something like this:
```nim
import std/dynlib

type
  NimibInitProc* = proc (doc: var NbDoc)

let lib = loadLib("plugin.so")
let initProc = cast[NimibInitProc](lib.symAddr("iniNimibPlugin"))
```

pietroppeter:
yep, that's pretty much the idea I had in mind, thanks for writing it down!

in general the idea is that:
[...]

PR #235, opened 2024-02-27
Refactor NbBlock

pietroppeter:
this is a major change, implements #168 (and also #117). Very WIP in a sandbox folder (see sandbox/notes.md)


pietroppeter:
milestone reached, now in sandbox/minib3.nim both a minimal html and json backend work. (your json work was great @HugoGranstrom). Still lots of details to work on, but I am feeling more confident now!

pietroppeter:
Some notes for me for later while I am on mobile related to NbRender type:
- right now NbDoc forced me to change the signature to add also Nb object
- I think I should change to pass only NbDoc object (type of the root object)
- also I could have NbRender support two types of functions (one for NbDoc like blocks one for NbText like blocks)
- I should encode the block type in the base block
- the general idea is to make the internal api safer (allow only what needs to be allowed)
- also the idea is to have the NbDoc the unit of serialization and rendering should not access anything outside of that (it will not be accessible from a deserialized version)
- and in NbDoc the plan is to have a theme field that is skipped on json serialization
- there is a tension between what you want to be able to serialize (content and specificities of a page) and what you want to be controlled (later) by the SSG (like the theme) and should not be serialized (also because it is redundant). Tension also with a third element which is the fact that rendering does need a theme
- as another, unrelated note, minimal theme could contain only doctype html and title element (see https://unplannedobsolescence.com/blog/best-hello-world-web-development/)

HugoGranstrom:
Wow! Great work! :partying_face: Looking good, it definitely seems way easier to define custom blocks now that the mustache backend is gone :smile: But I'm wondering how we will solve the problem of customizing an existing block type without it :thinking: But maybe that is what this idea is about?:

> - can I also use a super method or something to wrap the rendered block in a NbBlock?
>   - for example it could be used to add a class name inside a div
>     - this could be important for example to customize code block appearance
>   - or to add an optional id to a block
>   - (this could anyway be added later)
>   - and it could be added as something that by default a block does during rendering
>   - it might have a different signature (takes rendering of content and outputs new rendering)

This sounds magical! Being able to modify the rendered code would be so cool and useful. The custom block code could stay very simple while we add all sort of features behind the scenes. Would we be able to do this with raw HTML though? Parse the HTML to and AST and modify the AST and then rewrite it as HTML? 

> there is a tension between what you want to be able to serialize (content and specificities of a page) and what you want to be controlled (later) by the SSG (like the theme) and should not be serialized (also because it is redundant). Tension also with a third element which is the fact that rendering does need a theme

Is there really a tension? Couldn't we theoretically serialize the theme as well and just ignore/overwrite it in the SSG? 

Also, while we are refactoring the code, one part that has always confused me is `nb.blk`. I don't really see the point of it. Can't we just use `nb.blocks[^1]` instead (behind a template for easy access)? What value does adding it as its own variable (that you need to remember to set) bring?

pietroppeter:
note, I just pushed a possible implementation of a NbContainer object (and `NbDoc` would inherit from this):

```nim
  NbContainer = ref object of NbBlock
    blocks: seq[NbBlock]
    parent: NbContainer
```

while the html backend works fine with it the json backend fails to serialize because of the circular reference. This is an issue with `jsony` that does not support this out of the box. It does not seem easy to fix. I tried to use `skipHook` (with a dev version of jsony), but since `ref` objects delegate to the non ref version, it does not work (I do not know if there is a way to get the type of the non ref version). I also tried with a custom `dumpHook` (which would not be nice to automate) but it also does not work. This issues should be probably better documented (maybe in a jsony issue?).

My current plan is to actually skip the `parent` field in `NbContainer` and have a `containers: seq[NbContainer]` field in `Nb` object (and `add(nb: var Nb, blk: NbBlock)` as a generic api that might support a different implementation.

pietroppeter:
ok the new `NbContainer` implementation is indeed simpler, easier to implement and... it actually works! :)
I have also "hidden" the `nb.blk` api which is something that we might get rid of in the future.
 
Next steps:
- implement `nbCode` in minib
- implement `nbCodeFromJs` in minib
- think if there are essentially other types of blocks we need to test with the new refactoring
- if not go ahead and implement sugar to implement custom blocks (and refactor)
- clean up minib implementation
- reimplement all of minib starting with the new minib poc

HugoGranstrom:
Nice work! :star_struck: 

I really like the `withContainer` and `nb.add`. Now that `nb.blk` is hidden, I don't have that much against it anymore tbh.

HugoGranstrom:
So, I've started working on implementing `nbJsFromCode`. The part that I'm not sure about is where we should do the Nim->JS compilation. Should it happen before or after the transformation to JSON? It feels like it should happen before so that the JSON-HTML can happen without a Nim-compiler present + that steps become faster (important for SSG?). 

Another thing that I'm wondering about is whether we should do the compilation the same way we are doing it now. Now we do it in `nbSave`: [code](https://github.com/pietroppeter/nimib/blob/02b525313311c9ee8e74aa432af9cb89cbd4b3a0/src/nimib.nim#L265)
The thing is that we compile all of the `nbJsFromCode` blocks in the same file at the end, so we can't do the compilation before we have created all of them. I think the way we do it should be fine but I'm open to other ideas.

pietroppeter:
Agree it should happen before Json creation and it could be fine to have it in the nbSave node as before (I do not see at the moment advantages for the refactoring for this part).

HugoGranstrom:
Great, thanks for your input :D (does that sounds like an AI or is it just me? 🤣).

I agree, why change something that works. It is if/when we get plugin support (#209) that we could reconsider how we do things.

HugoGranstrom:
A first (hacky) implementation of `nbJsFromCode` is now implemented. Just have to add back a few fields to the NbDoc to make it work fully. The one problem is that we will probably need to define the types in `types.nim` so that they can be accessed in `jsutils.nim` without circular imports.

One thing I realized we will need now with container blocks is an iterator that walks through them all and returns a flattened list of blocks.

pietroppeter:
> we will probably need to define the types in types.nim

ok, for the moment I would probably only move the js block types there (maybe even in a jstypes), which do have a reason to be special, not all types definitions for blocks

HugoGranstrom:
Sounds like a good plan 👍

HugoGranstrom:
I've started thinking about the sugar for defining a block now and we have for example:
```nim
type
  NbImage = ref object of NbBlock
    url: string
template nbImage*(turl: string) =
  let blk = NbImage(url: turl, kind: "NbImage")
  nb.add blk
func nbImageToHtml*(blk: NbBlock, nb: Nb): string =
  let blk = blk.NbImage
  "<img src= '" & blk.url & "'>"
nbToHtml.funcs["NbImage"] = nbImageToHtml
addNbBlockToJson(NbImage)
```
Which should be shortened to:
```nim
newNbBlock(nbImage):
  url: string
  toHtml:
    "<img src= '" & blk.url & "'>"
```
I'm not sure how the template could be automated with this sugar. It feels like it should be created manually unless we can come up with a nice way of abstracting it. So currently I think this is feasible:
```nim
newNbBlock(nbImage):
  url: string
  toHtml:
    "<img src= '" & blk.url & "'>"

template nbImage*(turl: string) =
  let blk = NbImage(url: turl, kind: "NbImage")
  nb.add blk
```
Something else to consider is inheritance. By default the block inherits from `NbBlock` but you could also write:
```nim
newNbBlock(nbImage of NbContainer):
  ...
```
to signify that it should inherit another block?

HugoGranstrom:
Something we could do though to prevent the error of writing the wrong `kind` (wrong casing probably) in the template when creating the block, is to generate a `newNbImage` proc which has all the inputs except `kind` which is already prefilled.

pietroppeter:
Why the template cannot be created from the sugar macro? Is there some macro limitation I am not aware of (I know I am not aware of many things regarding macros... :))

pietroppeter:
Btw I forgot to add them but I think it might worth adding also a new `nb.` api that is not template based, e.g.:

```nim
proc image*(nb: var NbContext, url: string) =
  nb.blk.add NbImage(url: url)

template nbImage*(url: string) =
  nb.image(url)
```

The new procedural api would also help to reduce the number of templates that nimib needs and incidentally should reduce the number of globals (of course for `nbCode` and similar we still would have a template in order to have untyped arguments)

pietroppeter:
Also agree on the additional sugar for inheriting for other blocks

HugoGranstrom:
Because we don't know the inputs and what to do with them 😄 Say we have two fields on the object and we only want the user to provide one of them. So the problem is that we don't know what the input to the template should be from just the information given right now. We could of course add the template body as a new section in the sugar syntax with the inputs as well. But then we have basically written the template already. So I think it is more flexible if we write the template manually. For example when you have a block that you want to create a few different templates with different defaults for.

HugoGranstrom:
Yes it's a good idea to write the functional api as well 😄 I feel like I have regained some coding motivation now so I'll just need to find some time the coming weeks 😁

pietroppeter:
> Because we don't know the inputs and what to do with them

Ah right. I think we could provide maybe a default with all fields or indeed think of additional sugar but I guess it might be wiser for the moment to limit ourselves to leave it out of the sugar for now and see later if there is some good pattern that emerges

pietroppeter:
And yeah probably we should just provide the functional api as default (with all the fields) and leave all the templates as a customization (indeed nbCode has a lot of variants, the skip one, the block one, ... and they all refer to the same type)

pietroppeter:
In that case I would suggest the templates to use the functional api in their implementation (and this could be general advice).

HugoGranstrom:
> > Because we don't know the inputs and what to do with them
> 
> Ah right. I think we could provide maybe a default with all fields or indeed think of additional sugar but I guess it might be wiser for the moment to limit ourselves to leave it out of the sugar for now and see later if there is some good pattern that emerges

The problem with creating a default template is that it can cause conflict with a manually created one.

HugoGranstrom:
> And yeah probably we should just provide the functional api as default (with all the fields) and leave all the templates as a customization (indeed nbCode has a lot of variants, the skip one, the block one, ... and they all refer to the same type)

So we automatically generate the functional api? How would we do that for blocks that require untyped arguments? Or do you mean we just create a constructor automatically?

HugoGranstrom:
> In that case I would suggest the templates to use the functional api in their implementation (and this could be general advice).

Agreed, when possible, the templates should be sugar on top of the functional api 👍

HugoGranstrom:
Looking at this again, I'll try to summarize the last conversation we had:
- The sugar should generate a functional initializer with the `kind` prefilled
- The sugar should not generate the template (as we don't know what the inputs to it as they could be untyped)
- The sugar should not generate the functional API for the same reason, at least not by default. This is because it would create problems if we want to create a custom template/proc.

This means that for now, when defining a new block. We would have to create a functional api proc/template for each block type. And then create a "old-style" template using that (e.g. `nbCode` using `nb.code`)

This could be included in the sugar, but then we are basically defining the proc/template in there instead. Which could be a good thing as stuff would be nicely contained.

HugoGranstrom:
Sooo, now the sugar is finally finished. So instead of this:
```nim
type
  NbImage = ref object of NbBlock
    url: string
template nbImage*(turl: string) =
  let blk = NbImage(url: turl, kind: "NbImage")
  nb.add blk
func nbImageToHtml*(blk: NbBlock, nb: Nb): string =
  let blk = blk.NbImage
  "<img src= '" & blk.url & "'>"
nbToHtml.funcs["NbImage"] = nbImageToHtml
addNbBlockToJson(NbImage)
```
We now do this:
```nim
newNbBlock(nbImage):
  url: string
  toHtml:
    "<img src= '" & blk.url & "'>"

func image*(nb: var Nb, url: string) =
  let blk = newNbImage(url=url)
  nb.add blk

template nbImage*(url: string) =
  nb.image(url)
```
Note that `nbImage` must be a template as it needs to access `nb` which isn't defined until we run `nbInit`

HugoGranstrom:
I'm trying to think about what's left and it seems like I haven't converted nbJs to the sugar style yet. Will give it a try during the weekend. Other than that, is there anything else holding us back from actually starting to implement this in Nimib?

pietroppeter:
Hi! not really sure if there is anything missing, have not had really time to think about it. I guess we could try and start implementing in Nimib and see how it goes! Cheers from Brussels (Fosdem)!

HugoGranstrom:
Oh right it Fosdem! Have a great time and good luck! 😄 

Yes I'll finish nbJs up and give it a try and we'll see if I get caught up somewhere. Imagine if we got this working before 27th Febuary, only a year from opening the PR 😎

HugoGranstrom:
There's one thing I remember that we have discussed, but I can't remember what our conclusion was. Namely, target backends. The sugar has `toHtml` but do we want to support other formats as well in the future (like markdown) or are we putting our full focus on html? 

We have the `backend: NbRender` field on `Nb` so we can shift backends, but I have a hard time seeing how this arbitrary backend would work with our sugar syntax 🤔

pietroppeter:
Well, as long as we can still support custom backends, we do not need the sugar to support them. I guess the markdown backend could be an example on how to add a custom backend. I think it make sense to have the html backend as the only one to be supported by the sugar (although json is also kind of a backend of a special kind)

HugoGranstrom:
Okay, that sounds reasonable 👍 We could always expand the sugar in the future if the demand for it starts to exist (for example by adding a `toMd:` section in the sugar). But focusing on html for now sounds like a good idea 😄

HugoGranstrom:
And oh, backwards compatibility with the old mustache-backend must be handled somehow. It should be possible to implement as a custom backend during a transition period and then in the next major release we drop support for it? The things we'd like to remove are these fields from NbDoc and move them to the mustache-render object:
```
partials*: Table[string, string]
templateDirs*: seq[string]
renderPlans*: Table[string, seq[string]]
renderProcs*: Table[string, NbRenderProc]
```

HugoGranstrom:
Or should we just throw backwards-compatibility out the door and say that users can use Nimib 3.x if they want to use the old mustache backend? That would simplify things for us and we could make a more thorough refactoring of code now. It would mostly be libraries like nimibook and nimislides that would need to adapt their code to work with Nimib 4.0.

pietroppeter:
I would say to start implementing without thinking about backwards compatibility with old backend. Then at the end we can evaluate how much effort is to introduce some backward compatibility (with warning on usage and dropping on next version)

HugoGranstrom:
That sounds like a good plan 👍

HugoGranstrom:
I'm giving this a go again. Todays proudest achievement: 
```nim
func nbDocToHtml*(blk: NbBlock, nb: Nb): string =
  let blk = blk.NbDoc
  result = withNewlines:
    "<!DOCTYPE html>"
    """<html lang="en-us">"""
    headToHtml()
    "<body>"
    headerToHtml()
    leftToHtml()
    mainToHtml(blk, nb)
    rightToHtml()
    footerToHtml()
    "</body>"
```
A macro so we can more easily construct strings with newlines between them. And it also handles if-statements so we can conditonally include stuff as well. Compare this to the example in minib:
```nim
func nbDocToHtml*(blk: NbBlock, nb: Nb): string =
  "<!DOCTYPE html>\n" &
  "<title>" & blk.NbDoc.title & "</title>\n" &
  nbContainerToHtml(blk, nb)
```
So many fewer `&` and `\n`! 🥳

HugoGranstrom:
It is basically a templating macro, so I think we will do just fine without mustache. This is the most convoluted func I've written yet and it's not too bad:
```nim
func headerToHtml*(): string =
  result = withNewlines:
    "<header>"
    """<div class="nb-box">"""
    "<span>" & headerLeftToHtml() & "</span>"
    "<span>" & headerCenterToHtml() & "</span>"
    "<span>" & headerRightToHtml() & "</span>"
    "</div>"
    "<hr>"
    "</header>"
```

HugoGranstrom:
I totally forgot about string interpolation 🤣 So the usefulness of `withNewlines` isn't as remarkable as I though, BUT it is useful when we want to have if-statements (and for-loops when I implement support for them). Example where I could mix interpolation and macro:
```nim
func headToHtml*(doc: NbDoc): string =
  withNewlines:
    fmt"""
<head>"
  <title>{doc.getTitle}</title>
  <meta content="text/html; charset=utf-8" http-equiv="content-type">
  <meta content="width=device-width, initial-scale=1" name="viewport">
  <meta content="nimib {doc.context{"version"}.getStr}" name="generator">
  {doc.context{"stylesheet"}.getStr}
  {doc.context{"highlight"}.getStr}
  {doc.context{"nb_style"}.getStr}
  {doc.context{"latex"}.getStr}
    """
    if not doc.context{"disableHighlightJs"}.getBool(false):
      doc.context{"highlightJs"}.getStr
    "</head>"
```

HugoGranstrom:
Let the chaos begin! My objective is to get stuff work first, and then we can clean it up and move things around when we have something that works. The code now can run this simple example:
```nim
nbInit

nbText: "# Hello there"

nbSave
```
image

Now it's "just" a matter of rewriting all of the blocks to use the new code 🥳

pietroppeter:
Looking great, right now I have not much to add other than excited to see this come true :)

HugoGranstrom:
Thanks, glad to hear :D 

The only thing I'm feeling a little bit sad about is the fact that we don't have any way to customize the base theme easily. Previously you could do `nb.partials["footer_center"] = "<span>Footer center!</span>"`. But as I have written it now, it is a hardcoded function that returns an empty string because that's what we had before. A user can't override that as its hardcoded. 

So I'm thinking about keeping partials but making them functions instead of strings. So basically:
```nim
type
  NbPartialFunc* = proc (doc: NbDoc, nb: Nb): string 
  Nb = object
    ...
    partials*: Table[string, NbPartialFunc]
```
The question then is if it should be part of the renderer/backend or if it should be part of the `Nb` object. It feels generic enough that it could be placed in `Nb`, but at the same time it feels like it belongs in the renderer so different renderers can do it differently.

HugoGranstrom:
Todays progress:
- Removed need for inputing command to `codeAsInSource` so now we have a neat `getCode` template that you can simply use wherever we have a `body` and want its code.
- Almost all blocks have been ported to the new system 🥳  Just a few left more left...

HugoGranstrom:
I couldn't resist finishing up the last few blocks as well 🤭 We are so close now! The things left on my TODO is:
- [x] Fix merge conflicts
- [x] Fix docs - they are using some old features we no longer support
- [x] Think about a partials system for the HTML backend to make it more flexible
- [x] Investigate how these changes will affect nimiSlides & nimiBook to see if we are missing any features
- [x] Fix tests / rewrite parts from scratch
- [x] Clean up the code

HugoGranstrom:
The first few pages of the docs are building! 🥳 https://eee9efe34998c4e01600--nimib.netlify.app/

pietroppeter:
Regarding the partial/backend thing, this is definitely something I am noting myself down to try and help thinking out a solution. This is definitely a breaking change, but I think it should be worth it.

Not sure if it helps but the general idea in the past was: a backend is a way to compile to a different format (html vs md vs ...). Backend could support theming and overriding (but basically it made sense for html theme only.

Right now we kind of realized that maybe backend abstraction is less customizable than we thought (json backend needs to be baked in). And we could try and keep only a layer of customization at the theme level. Not necessarily it needs to work in the same way (i.e. we could have a way of overriding by inheriting...).

Still if, after thinking of what should be the best way to support theming, we figure out also a way to support the old way (mustache had the advantage that you could also change the theme with files that override partials) it would be nice.

These are my immediate thoughts and I might be missing a lot. Probably at some point it could be nice to have a lice conversation about this, but maybe in the meantime we could start an async one.

HugoGranstrom:
Thanks for your thoughts 😄 Yes it is a bit of a complicated matter. I still think the backend+ theme is a good way to work. But we need to be a bit more precise in exactly what each part should be responsible for. They way we did it before the refactor was:
- The theme populated the context object with global variables and settings
- The backend handled all things rendering using the values in the context.
So here the theme wasn't really about how the blocks were rendered (well some blocks did use context values but let's skip them for simplicity). It mostly controlled the CSS for theming. 

One approach going forward I could see is moving the customizable partials to the theme as well. And the backend's responsibility is to render the block from those partials and hardcoded HTML. Having written this though, I don't really see the point in separating them. A theme will generally only work with one backend, so why split them up? Let the theme decide the backend. Or am I missing something here?

HugoGranstrom:
Thinking about it a bit more, I see two kinds of customizations (using nimiSlides as reference)

1. Page-wide customization. Typically, the document layout and CSS (nimiSlides uses Reveal.js and need another page layout than base nimib).
2. Block-rendering-customizations. Typically, changing how a certain block looks. (nimiSlides adds classes to `nbCode` that Reveal.js needs)

The first kind of customization is currently possible by modifying `nbToHtml["NbDoc"] = renderNimiSlidesDoc`. The second kind of customization is also possible in the same way `nbToHtml["NbCode"] = renderNimiSlidesNbCode`. BUT I would have to do this for all NbCode-like blocks because there is no way to have reusable and exchangable parts in them. For example:

```nim
toHtml:
    withNewlines:
      if blk.code.len > 0:
        &"<pre><code class=\"nohighlight hljs nim\">{blk.code.highlightNim}</code></pre>"
      nbContainerToHtml(blk, nb)
      if blk.output.len > 0:
        &"<pre class=\"nb-output\">{blk.output}</pre>"
```
Even if I rewrote this to use functions for rendering the code and output, it wouldn't be possible to change it because the function is hardcoded. So I would have to rewrite the function altogether and replace the function calls with my own. So if instead I could do:
```nim
toHtml:
    withNewlines:
      nb.renderPartial(blk, "nbCodeSource")
      nbContainerToHtml(blk, nb)
      nb.renderPartial(blk, "nbCodeOutput")
```
Then I would be able to change the partials and it would change everywhere it is used (in all NbCode-like blocks). BUT the problem is that these two particular partials would need `blk` to be a child-type of `NbCode` which is hard to enforce. In this case it is fair to require it. But say that we have a more general partial that in theory would work with all types that has an `output` field. How would we handle that? One idea is to convert all blocks sent to a partial to JsonNode. That way we could pass all sorts of blocks to the partial and it would work. 

One thing I've realized when writing this is that we aren't really using the inheritance a lot at all when rendering. We are using it for storage and creation. But it feel like we have untapped potential when it comes to rendering. I guess it is hard when you have different backends and want customization. Then we are basically stuck with the table-of-functions that we are using now. So maybe it isn't even desirable to use inherticance when rendering 🤔 

This was a lot of rambling and I mostly wrote it down to not forget it hehe😅So no need to reply unless you found something useful.

HugoGranstrom:
all docs are up! (except index.md because of lack of md backend) 🎊

HugoGranstrom:
I've been thinking a bit more, and I think it boils down to that we want a system that is:
1. Customizable
  - It should be possible to change how a block is rendered
2. Re-usable
  - It should be possible to re-use components from other blocks. For example, I should be able to style my block like an `nbCode` block.
  - BUT, it should ideally be possible without my block being a direct descendant of the block. Imagine I want to use parts from multiple different blocks, for example. My block can't inherit from both of them!

Our options thus far:
- `Table[string, proc(blk: NbBlock, nb: Nb): string`
  - Fulfills requirement 1 but not requirement 2
  - The problem stems from the fact that the `NbBlock` must be converted to the expected kind to access its fields. This only works for direct descendents of the block.
- Inheritance
  - Doesn't fulfill any of the requirements 
- Regular functions
  - Fulfills requirement 2 but not requirement 1
- `Table[string, proc(blk: JsonNode, nb: Nb): string`
  - Fulfills both requirements
  - The fact that it accepts a `JsonNode` allows us to pass in different parameters to different procs
  - This would also allow us to pass in other parameters than the fields of the blocks. For example, we could pass in a list of classes to a block. 
  - The downside is that we get fewer compile-time guarantees and have to do more manual checks to see if fields exist. But hopefully this can be dampened by sugar. 

I may have missed some options, but currently I see embracing JsonNodes as our best bet. So a typical `toHtml` proc could look like:

```nim
func somePartial*(blk: JsonNode, nb: Nb): string =
    if not blk{"code"}.isNil:
      let classes = blk{"classes"}.getElems.mapIt(it.getStr).join(" ")
      &"<code class=\"{classes}\">" & blk{"code"}.getStr & "</code>"
    else:
      ""
nb.partials["somePartial"] = somePartial

func someBlockToHtml*(blk: NbBlock, nb: Nb): string =
  let blk = blk.SomeBlock
  var jObj = blk.toJson
  jObj["classes"] = %["highlightjs-nim", "reveal-js-code"]
  withNewlines:
    "<h1>" & blk.title & "<h1>"
    nb.renderPartial("somePartial", jObj)
```

It is more verbose, but I think we could make sugar to simplify the process quite a bit.

What do you think?

HugoGranstrom:
The biggest problem with this, I would say is that it is hard to know what fields a particular partial function expects. I don't see how we could handle that at compile-time. But some sugar to check for required and optional fields at runtime should be possible. Then the writer of partials would have to write fewer checks manually at least.

pietroppeter:
This is very interesting development, will have to think and reconnect with the actual code to give an informed opinion (current goal is to find time this weekend).

Just so I understand, right now we have been implementing inheritance part (and part of the complexity of serializing to json came from that), but according to what you are saying, it is not worth the effort and we should instead revert to something more similar to what we had (only going all in on Json Node)?

pietroppeter:
First thought anyway would be to use a `distinct JsonNode` to control the api (but this of course is a minor point).

HugoGranstrom:
I think we should keep inheritance (different field for different blocks are good and we get some kind of CT-safety). But for the rendering we don't have much use for it. So there we covert it to JsonNode. But going all in on distinct JsonNode is also an option I haven't considered :O

pietroppeter:
Ah ok I was understanding wrong. :) Yeah, all in to JsonNode for rendering looks very reasonable for me.

HugoGranstrom:
Nice to hear. I'll give a try implementing it today and post an example here so we can discuss if we like the syntax and what parts we would like sugarify 😄

HugoGranstrom:
It was surprisingly easy to implement! And the best thing about this partials stuff is that it's totally optional. If you (a random nimib block creator) don't want to use partials, you can just go your merry way and don't even have to deal with `JsonNode`s. 

Here's how we implement `NbCode` now:
```nim
func nbCodeSourcePartial*(blk: JsonNode, nb: Nb): string =
  let code = blk{"code"}.getStr
  if code.len > 0:
    &"<pre><code class=\"nohighlight hljs nim\">{code.highlightNim}</code></pre>"
  else:
    ""

nbToHtml.partials["nbCodeSource"] = nbCodeSourcePartial

func nbCodeOutputPartial*(blk: JsonNode, nb: Nb): string =
  let output = blk{"output"}.getStr
  if output.len > 0:
    &"<pre class=\"nb-output\">{output}</pre>"
  else:
    ""

nbToHtml.partials["nbCodeOutput"] = nbCodeOutputPartial

newNbBlock(NbCode of NbContainer):
  code: string
  output: string
  toHtml:
    withNewlines:
      nb.renderPartial("nbCodeSource", jsonutils.toJson(blk))
      nbContainerToHtml(blk, nb)
      nb.renderPartial("nbCodeOutput", jsonutils.toJson(blk))
```
It is (as you said above), basically our old mustache-workflow, but with `JsonNode` instead.

Here's a proposal for a sugar:
```nim
func nbCodeSource*(nb: Nb, text: string = defaultValue, code: string): string {.nimibPartial.} =
	text & code

# rewritten to:
func nbCodeSource*(nb: Nb, blk: JsonNode): string =
	let text = blk{"text"}.getStr(defaultValue)
	let code = blk{"text"}.getStr(default(string))
	text & code
```
It should be pretty straight forward to implement. It helps document the inputs as well as they are part of source code. The handling of the abscence of a certain fields could probably be improved. For example, if no default value is given, maybe we raise an exception as it's a mandatory field? Then the `let code = ...` line would first check that the key exists and is a string. 

The problematic part is `seq[T]`, `Table[string, T]`, as it involves nesting. But it's not undoable so we'll handle it in due time.

pietroppeter:
Looks good to me!
I like also the sugar.

Couple of things I am noticing:
- we are inheriting nbCode from NbContainer to allow for container blocks in nbCode, right? (so that we do not need anymore nimibCode)
- TIL about the [braces syntax in JsonNode](https://nim-lang.org/docs/json.html#%7B%7D%2CJsonNode%2Cvarargs%5Bint%5D) I guess the advantage is it does not raise KeyError, right?

HugoGranstrom:
Nice! Then I'll go ahead and implement it for the other blocks as well 👍 

> we are inheriting nbCode from NbContainer to allow for container blocks in nbCode, right? (so that we do not need anymore nimibCode)

Exactly, `nimibCode` is basically an alias for `nbCode` now.

> TIL about the [braces syntax in JsonNode](https://nim-lang.org/docs/json.html#%7B%7D%2CJsonNode%2Cvarargs%5Bint%5D) I guess the advantage is it does not raise KeyError, right?

That is correct. I also didn't know about it until a few weeks ago. It's really nice as it handles `nil` sielently so you can chain multiple of them without worrying:
```nim
var j: JsonNode = nil
j{"hello"}{"world"}.getStr("default string")
```
This will just work, `j{"hello"}{"world"}` evaluates to `nil` and `.getStr("default string")` checks if it is nil or of the wrong kind and then uses the default value. So it is already really ergonomic to work with `JsonNode`s :D

Edit: it can be written even shorter:
```nim
j{"hello", "world"}.getStr("default string")
```

HugoGranstrom:
https://github.com/HugoGranstrom/nimiSlides/pull/52 
I think this PR shows that it is totally possible to rewrite a nimib library using this refactoring.

HugoGranstrom:
I can't imagine nimibook being more complicated. It doesn't really define that many custom blocks, it's more of the tooling around it and a theme. So I would be okay with saying that this PR doesn't break anything for the other libraries. And that means (by looking at my todo list hidden in the middle of this thread) that I should be fixing a rewriting tests 🥳

HugoGranstrom:
The tests are finally fixed! :partying_face: 
It seems like the test run for Nim 1.6 is failing for mysterious reasons, but stable is working!
And the preview deploy works as well! https://f0285381d1083ae4eeb2--nimib.netlify.app/

We are so close, now we just need to go through the code and clean it up. One thing I'm thinking about is moving all the block definitions out to other files. Maybe a folder with different blocks in different files? Maybe just a gigantic file with all the blocks? Do you have any preferences? The drawback of putting them in different files would be that we could stumble upon cyclic imports if we want to share stuff between them.

pietroppeter:
I would probably go with single file with all blocks

HugoGranstrom:
Great, then I'll fix that when I get some time over 👌

HugoGranstrom:
That worked out way smoother than I expected. Now I just need to clean up some comments and then we should be ready for a final review. :partying_face:

HugoGranstrom:
There was less to be done than I thought, so now the PR is officially ready for review @pietroppeter :wink:

pietroppeter:
Great will give it a go soon enough. You should mark it as non draft

pietroppeter:
> **Note: This is a Claude-generated code review, produced in a [Claude Code Web session](https://claude.ai/code/session_016BUPXppnTex6HQuS2CAVzK).**

## Review: PR #235 — Refactor NbBlock

### Summary of Changes

#### 1. New wrapper type `Nb` (previously `nb: NbDoc`)

The old `nb` was an `NbDoc` value directly. Now `nb` is an `Nb` wrapper object containing `nb.doc`, `nb.blk`, `nb.containers` (stack for nested blocks), and `nb.backend`. This separation keeps document data (`NbDoc`) cleanly distinct from the build-time state needed to produce it (`Nb`) and enables the `withContainer` pattern.

**Breaking change:** code that accessed `nb.source`, `nb.filename`, `nb.thisFile`, `nb.options`, `nb.cfg` etc. directly now needs `nb.doc.<field>`. End-user notebooks using `nbCode:`, `nbText:` etc. are unaffected.

---

#### 2. `NbBlock` becomes a polymorphic ref hierarchy

The old `NbBlock` had `command`, `code`, `output`, `context` as flat fields on every block. Now each block type is its own `ref object of NbBlock` with only the fields it needs. The `kind: string` field on the base type carries the type name for JSON dispatch. New types: `NbCode`, `NbText`, `NbTextWithCode`, `NbImage`, `NbFile`, `NbVideo`, `NbAudio`, `NbDiv`, `NbRawHtml`, `NbJsFromCode`, `NbJsFromCodeOwnFile`, `NbPython`.

---

#### 3. Mustache dependency removed

`requires "mustache >= ..."` removed from `nimib.nimble`. Rendering is now pure Nim functions — type-safe, no external template language for theme authors to learn. (The docsdeps task still installs `mustache@0.4.3` for example docs that use it directly — that's fine.)

---

#### 4. New `NbRender` system (`globals.nim`, `renders.nim`, `themes.nim`)

`nbToHtml` and `nbToMd` are `NbRender` ref objects each holding:
- `funcs: Table[string, NbRenderFunc]` — per-block-kind render function keyed by type name
- `partials: Table[string, NbPartialFunc]` — named fragments (head, header, footer, nbCodeSource, etc.) that themes override

Themes configure the active backend by swapping partials and setting `nb.doc.context` JSON values.

---

#### 5. `newNbBlock` macro and `nimibSugars.nim`

Eliminates boilerplate: one `newNbBlock(NbFoo): field: type; toHtml: ...` call auto-generates the `ref object` type, a `newNbFoo(...)` initializer, the `nbFooToHtml` proc, registration in `nbToHtml.funcs`, and `addNbBlockToJson(NbFoo)`. The `withNewlines` macro handles the common pattern of building multi-line HTML strings conditionally.

---

#### 6. All block definitions moved to `builtinBlocks.nim`

`src/nimib.nim` goes from ~300 lines to ~70 lines. All block definitions live in `builtinBlocks.nim`. JS/Python blocks live in `jsutils.nim`.

---

#### 7. JSON serialization (`jsons.nim`)

Enables document interchange and is a prerequisite for nimibook. `addNbBlockToJson` registers parse/dump hooks for each block type via the `kind` string discriminator using `jsony`.

---

#### 8. `NbContainer` and nested blocks

`NbContainer` adds `blocks: seq[NbBlock]`. `nb.withContainer(blk): body` manages the stack. `blocksFlattened` provides a flat traversal.

---

### Items Worth Addressing

#### A. Dead code: `newNbBlockOld` in `blocks.nim` 🔴
Renamed from `newNbBlock` with a `# do we need this anymore?` comment but never removed. It still references `newContext` (mustache). Since mustache is no longer a dependency this will fail to compile if called. Should be deleted.

#### B. `nbJsCounter` and `id` are in `NbDoc`, not `Nb` 🟡
The comment in `types.nim` explicitly says `# Nb should just contain stuff needed for producing the NbDoc (like id and nbJsCounter)` — but both fields are still on `NbDoc`. These are build-time counters, not document content. Intent not implemented yet.

#### C. `NbDoc.context` is still a `JsonNode` grab-bag 🟡
Theme config values (stylesheet URL, title, highlight settings, github URL, `path_to_root`, `source`, etc.) all live in `nb.doc.context: JsonNode`. This is a deliberate compromise (typed fields for block content, JSON for theme/metadata) but the boundary is fuzzy. Worth a comment documenting the design decision.

#### D. Version number not bumped 🔴
`.nimble` still reads `version = "0.3.13"`. The `nb: NbDoc` → `nb: Nb` change is a breaking API change for anyone accessing `nb` fields directly — this warrants at least `0.4.0`.

#### E. No changelog entry 🟡
`changelog.md` has no entry for this refactor. At minimum the "Breaking changes" section should describe the wrapper change for advanced users.

#### F. `NbJsFromCode` should inherit from `NbCode` but doesn't 🟢
The comment in `jsutils.nim`: `# should inherit from nbCode, but it is defined in nimib.nim...`. Now that all built-in blocks are in `builtinBlocks.nim`, this circular dependency is gone. The inheritance could be done now.

#### G. `withNewlines` doesn't handle `for` loops 🟡
`nimibSugars.nim`: `# TODO: handle for loops` — `withNewlines` silently produces wrong output if the body contains a `for` loop. Could be surprising for theme authors writing custom renders.

#### H. `NbFile` render is not using partials 🟢
The comment in `builtinBlocks.nim`: `# Make pre-code with hljs-extensions into a partial!` — inconsistent with how `NbCode` renders (which does use partials). Resolve or remove the comment.

#### I. `nbInit` parameter renamed `backend` → `renderer` 🟢
Minor unnecessary churn for anyone using the named parameter form `nbInit(backend = ...)`.

---

### Overall

The refactoring is architecturally sound. The core changes (polymorphic blocks, mustache removal, `newNbBlock` macro, centralised `builtinBlocks.nim`) work together coherently and accomplish the goals described in #168 and #117.

Must-fix before merging: **A** (dead code referencing removed mustache dependency) and **D** (version bump). Items B, E, F, G, H are good candidates for follow-up issues.

pietroppeter:
Tried to review with Claude. It does seem it has done a decent job, but now of course, I need to review the review :)

HugoGranstrom:
> B. nbJsCounter and id are in NbDoc, not Nb 🟡

I'm thinking a bit about this and it doesn't hurt to include these fields in the `NbDoc` instead of in `Nb`. For example if we want a feature in the future where you can append to an already existing NbDoc (loaded from a json file). Then it's good to know what the id counter is at to avoid collisions. So I vote for keeping it as is and just remove that comment. What do you think?

HugoGranstrom:
> F. NbJsFromCode should inherit from NbCode but doesn't 🟢

I tried fixing this, but we must have a bigger cycle of imports that I can't seem to find. So I'll keep it as it is for now

pietroppeter:
honestly I think the review was not bad, the open points in the review have been addressed, I should have had a look earlier, but this seems good to me to merge, do you want to do the honors @HugoGranstrom and merge? if people start using nimislide with the new version for nimconf we might receive feedback/issues and iron up stuff.

pietroppeter:
only remark I could find (really minimal) is that `preCodeTag` could be used also in `nbCodeSourcePartial` adding a highlight deafult parameter (default value highlight = true?) https://github.com/pietroppeter/nimib/blob/08238e4078c42a757fea200b912fca439f4ba502/src/nimib/renders.nim#L28

pietroppeter:
see a recent CI failure (not due to our code I think, but could be fixed here) https://github.com/pietroppeter/nimib/pull/252#issuecomment-4397475904

HugoGranstrom:
Good idea with `preCodeTag` :D When the tests are green, I will do the magical merge :magic_wand:

HugoGranstrom:
Here we goooooo!

Issue #62, opened 2021-07-07
Cannot open pathutils


cvanelteren:
Hi,

When trying to compile my first book (very exiciting!) I am getting the following error

``` Error: cannot open file: /usr/compiler/pathutils```

I am running arch. Judging from other post on the forum and nim git may be the underlying issue. How do I deal with this bug?
Best,
C

**Comment by HugoGranstrom** [comment]
That's odd, have never seen that error before. 🤔 Which version of Nim are you running?

**Comment by pietroppeter** [comment]
[`pathutils`](https://nim-lang.org/docs/compiler/pathutils.html) is a stdlib module with [a non evident way to import it](https://forum.nim-lang.org/t/7190) and nimib uses it.

Can you try to compile a nim file with only:

```nim
import "$nim/compiler/pathutils"
```

And also one with only:

```nim
import ../compiler/pathutils
```

Nimib uses the first way to import it.

And also as @HugoGranstrom metions, report nim version (full output of `nim --version`).

cvanelteren:
I am using

```nim
Nim Compiler Version 1.4.8 [Linux: amd64]
Compiled at 2021-06-07
Copyright (c) 2006-2021 by Andreas Rumpf

active boot switches: -d:release -d:nativeStackTrace
```

pietroppeter:
I found only this possibly related issue for Archlinux: https://github.com/nim-lang/Nim/issues/14040. My guess is that  `"$nim/compiler/pathutils"` might not work for archlinux, maybe `../compiler/pathutils` might work (maybe requiring to add a specific path). If you find that there is a workaround which works for you with the second version, I can change how nimib imports pathutils. Anyway it seems to me something that should be solved upstream in nim and we might want to open an issue in nim main repo.

Regarding the workaround with the second version, you could probably check if you arch nim distribution comes with a `pathutils.nim` somewhere. If you have it and it is in a `compiler` folder, then adding to the path the `compiler` folder itself should work.

pietroppeter:
[...]

PR #238, opened 2024-03-15
paths.nim: Fix import for FreeBSD

lbartoletti:
`compiler` doesn't seem to be standard.
At least, in FreeBSD, it's in ${LOCALBASE}/nim/lib/compiler where ${LOCALBASE} is /usr/local by default.

Add a when defined to import `pathutils` on FreeBSD.

### DIFF
--- a/src/nimib/paths.nim
+++ b/src/nimib/paths.nim
@@ -1,4 +1,7 @@
-import "$nim/compiler/pathutils"
+when defined(FreeBSD):
+  import "$nim/lib/compiler/pathutils"
+else:
+  import "$nim/compiler/pathutils"
 export pathutils
 import os, strutils, hashes
 


pietroppeter:
Thanks for this, there is indeed a related issue https://github.com/pietroppeter/nimib/issues/62 that I thought it was only arch specific but it turns out it might affect also freeBSD. I am not planning to test this but it looks a reasonable change and hopefully it helps on your side.

Looks ok for me, if @HugoGranstrom cannot come up with objections I would merge it.

HugoGranstrom:
Yes looks good to me as well 👌

lbartoletti:
> hopefully it helps on your side

Yes, it helps me and I'm starting to write my blog, and documentation with it :)

And... thanks to you and hugo for these tools!

PR #239, opened 2024-03-23
Update changelog.md with 3.11 release notes and bump 3.12

pietroppeter:


pietroppeter:
forgot to do that the other day, now that this is done we are ready to release again any time we have a new fix/feature :)

PR #242, opened 2024-06-16
Make Nimib logging to stdout optional (`-d:nimibQuiet`)

neroist:
I'd rather not have Nimib print to stdout all the time & fill up the terminal (esp. when compiling multiple documents!), and its better for users to have the option to disable logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ tracks further options in `nb.options.other: seq[tuple[kind: CmdLineKind; name, 
 
 nimib's behavior can be further turned via Nim's define flags:
 
-* `-d:nimibNoLog`: Completely disables nimib's logging to stdout
+* `-d:nimibQuiet`: Completely disables nimib's logging to stdout
 * `-d:nimibCodeFromAst`: Makes nimib capture block code from AST of body (as opposed to from file source; see next section). Available since version 0.3
 
 ### Code capture
--- a/README.md
+++ b/README.md
@@ -257,6 +257,13 @@ Nimib options:
 The value of options are available in `nb.options` field which also
 tracks further options in `nb.options.other: seq[tuple[kind: CmdLineKind; name, value: string]]`.
 
+### define flags
+
+nimib's behavior can be further turned via Nim's define flags:
+
+* `-d:nimibNoLog`: Completely disables nimib's logging to stdout
+* `-d:nimibCodeFromAst`: Makes nimib capture block code from AST of body (as opposed to from file source; see next section). Available since version 0.3
+
 ### Code capture
 
 The code capture of a block like `nbCode` (or other custom blocks)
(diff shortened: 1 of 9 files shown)


HugoGranstrom:
This option makes a lot of sense to add 👍

neroist:
Another issue I have is that Nimib changes the current directory, which forces me to change it back myself (I use Prologue which searches for `./.config`). I'll make an issue when/if I'm able get the wording right, and PR if I'm able to fix it.

pietroppeter (approved):
Looks great to me! (the only moment I paused is the echo left in blocks.nim, but I get that since we have a different type of output "prefix" we cannot reuse log and leaving the echo seems the most appropriate thing)

pietroppeter:
Thanks for this PR, it definitely makes sense to add the flag!

This also will help the moment we want to improve logging in some way (but not doing that before someone has a clear use case or a need ;)).

@HugoGranstrom not sure if you also want to review, otherwise for me this is good to merge!

pietroppeter:
> Another issue I have is that Nimib changes the current directory, which forces me to change it back myself (I use Prologue which searches for `./.config`). I'll make an issue when/if I'm able get the wording right, and PR if I'm able to fix it.

yes, there is a reason it changes directory (to access/create assets in the "build" directory, actually called homeDir, so that if I generate an image it is part of the assets that the static site will be able to see), but I can see other use cases not wanting that. I think it make sense to add a flag there (we are starting to have a bunch of them, at some point we might want to document them somewhere).

For reference, the change of directory is here: https://github.com/pietroppeter/nimib/blob/02b525313311c9ee8e74aa432af9cb89cbd4b3a0/src/nimib.nim#L46

pietroppeter:
...and I just realized that this feature (`-d:nimibLog`) should be documented somewhere (I guess in the Readme/nimib main doc), otherwise it is only for the insiders :)

HugoGranstrom:
A bit of documentation and I'm content with merging this 😁🚀

neroist:
looking at [`docsrc/config.nims`](https://github.com/pietroppeter/nimib/blob/02b525313311c9ee8e74aa432af9cb89cbd4b3a0/docsrc/config.nims#L2C3-L2C16) I realize that `nimibQuiet` is probably a better & more elegant name than `nimibNoLog` 😅

HugoGranstrom:
> looking at [`docsrc/config.nims`](https://github.com/pietroppeter/nimib/blob/02b525313311c9ee8e74aa432af9cb89cbd4b3a0/docsrc/config.nims#L2C3-L2C16) I realize that `nimibQuiet` is probably a better & more elegant name than `nimibNoLog` 😅 

Oh look at that :o Yes `nimibQuiet` would probably be a better name for the switch

neroist:
name changed & docs added! 👍

pietroppeter:
haha, forgot about that `nimibQuiet`, I must have introduced it back then for the same purpose :)
I am a bit on the fence on the change of name since `nimibNoLog` was really clear in its purpose, but `nimibQuiet` is more playful and you both seem to approve of it.

thanks again, looks good, will edit the title (so that when we make the changelog it will be already correct) and will merge this.

PR #153, opened 2022-11-13
add nbVideo and nbAudio

pietroppeter:
(I had this branch sitting around, I guess I will have to finish this at some point...)

PR #244, opened 2024-06-26
Add nbVideo & nbAudio (revive #153)

neroist:
for `all_blocks.nim` I wasn't sure what to put, so I ended up just putting Bad Apple instead :P please feel free to change! maybe reference some of that stuff in the [Q&A](https://github.com/pietroppeter/nimib?tab=readme-ov-file#--q--a). or maybe you should keep it instead, perhaps.

Some changes made:
* `relToRoot` uses [`isAbsolute`](https://nim-lang.org/docs/uri.html#isAbsolute%2CUri) -- thought this was better than checking for "http" as the first four letters of the string
* You can now set the MIME type of the media (doesn't assume based on file ext.), and set the `autoplay`, `muted`, and `loop` attributes of the video/audio html tag


pietroppeter:
thanks! the changes greatly improve on #153 :)

I just learned about bad apple [yesterday] and I am definitely very amused 🤣  by this coincidence 🤯 and approve of the choice 🍎. 

Instead of references in Q&A we could:
- add a link to this [explainer] as the example url for `nbVideo` element (two advantages: 1) we show the url can be both relative and a web url; 2) we can remove the 7MB video asset)
- invert `nbAudio` and `nbVideo` section in `allblocks.nim` (so that the audio comes before the "explainer")

Last thing is to ask if @HugoGranstrom did know about bad apple (he is definitely younger than me) and if he is ok with the choice.

[yesterday]: https://x.com/pietroppeter/status/1806868565259059284
[explainer]: https://youtu.be/6QY4ekac1_Q?si=qCOOr8KIO3ffxMXQ

HugoGranstrom:
Had no idea about Bad Apple, but I definitely applaude the choice 😂

neroist:
> add a link to this [explainer](https://youtu.be/6QY4ekac1_Q?si=qCOOr8KIO3ffxMXQ) as the example url for nbVideo element

afaik, its impossible to insert a Youtube video into a `<video>` tag without some external Javascript dependency. We should probably add a link to the explainer in the nbVideo section instead.

For the audio & video assets, those can be replaced by either of [these](https://archive.org/details/MasayoshiMinoshimaFeat.Nomico-BadApple) [two](https://archive.org/details/TouhouBadApple) from archive.org.

pietroppeter:
Ah I guess you are right. This poses the question, should nbVideo be just a little wrapper for html video tag or also do the right thing (to embed a YouTube video we could use a iframe instead of JS right?) in case eg you put a link to YouTube?

I do not think we should get an answer in this PR, we could open a new issue if we feel like discussing this.

For the scope of the PR I would just add a link to the explainer for the moment. I think assets of a few MB are fine and there is no need to go with archive (which I guess might be slower to be reached? Or maybe not?)

neroist:
sorry it took so long! should be ready to merge, no?

pietroppeter:
yep, looks good to me, thanks!

PR #246, opened 2024-06-29
Create homeDir if it doesn't exist alredy

neroist:
Title is self-explanitory -- unsure if the directory's creation should be logged or not (probably not?)


pietroppeter:
yep, self explanatory and straightforward fix to a "bug" we did not know about. Thanks!

Looking at this https://github.com/pietroppeter/nimib/blob/6daf48952f18384fcfae679d3dce5a77625fcfe1/src/nimib/docs.nim#L9
I would say for consistency to log directory creation. (and I think in general looks a good thing to log by default changes to the environment, since we are defaulting for now to a rather verbose nimib).

pietroppeter:
Thanks, looks good!

PR #250, opened 2024-11-18
Fix CI: upload-artifacts

HugoGranstrom:
It's due time we try to fix the preview in the CI

### DIFF
--- a/.github/workflows/netlify_build_docs.yml
+++ b/.github/workflows/netlify_build_docs.yml
@@ -19,7 +19,7 @@ jobs:
       - name: build docs
         run: nimble docs
       - run: echo Commit hash = ${{ github.event.pull_request.head.sha }}
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           name: build-${{ github.event.pull_request.head.sha }}
           path: docs/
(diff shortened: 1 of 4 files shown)


HugoGranstrom:
Strange, I can't reproduce the error locally :/

HugoGranstrom:
@quimt do you have any ideas what might be going on?

quimt:
> @quimt do you have any ideas what might be going on?

I don't know in detail, but happyx does use macros to produce [emit pragmas](https://nim-lang.org/docs/manual.html#implementation-specific-pragmas-emit-pragma). I haven't encountered this bug using happyx by itself, so I suspect it has something to do with how happyx interacts with nimib's protocol for sandboxing the code. Whatever the reason, emiting something other than a string doesn't make sense as far as I know.

HugoGranstrom:
> I don't know in detail, but happyx does use macros to produce [emit pragmas](https://nim-lang.org/docs/manual.html#implementation-specific-pragmas-emit-pragma). I haven't encountered this bug using happyx by itself, so I suspect it has something to do with how happyx interacts with nimib's protocol for sandboxing the code. Whatever the reason, emiting something other than a string doesn't make sense as far as I know.

Thanks for the leads :D Judging from the error message and this code: https://github.com/HapticX/happyx/blob/3b1d08aaed757cd1747c50a5f1d480444fa18b20/src/happyx/private/macro_utils.nim#L1065
It seems like whatever `args[1][0]` is, it isn't what is't supposed to be. I'll continue digging.

HugoGranstrom:
Ahh, the changes I make to the CI files aren't run... So I will have to merge the changes and just hope it works and then test it with a new PR 🙈

PR #252, opened 2026-05-07
Update Nimib conference presentation links (add 2024, fix 2021) - and NimibLand migration node

pietroppeter:
Updated conference presentation details for Nimib, adding link for NimConf2024 and fixing link for NimConf2021. I think NimConf 2022 presentation still deserves the main place with thumbnail.


pietroppeter:
added also a small note regarding the never ending migration to NimibLand

pietroppeter:
and CI is red (seems like re module is the culprit)!
image

we probably could still merge this inessential doc change and fix this in #235 ?

HugoGranstrom (approved):
Looks good to me :D

PR #253, opened 2026-05-25
Changelog and small fixes before 0.4.0

HugoGranstrom:
@pietroppeter Here is the changelog and some small renames (`tname` -> `name` now that we aren't using templates everywhere). I would appreciate a readthrough of the changelog and when that is done we are ready for an official release! :rocket:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,9 +13,103 @@ Notes for maintainers:
 - When we tag a new release, we should auto generate the release notes. It does not hurt if we add more context to the release notes (e.g. taking notable elements from PR discussion). We might also want to add a release discussion post.
 - finally, after a release, we update this changelog (and bump version) using the same wording from release notes: https://github.com/pietroppeter/nimib/releases
 
-## v 0.4.0
+## v0.4.0
+### Changes in how to define blocks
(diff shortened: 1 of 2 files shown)


pietroppeter:
Nice changelog I think it makes  very good work in explaining the differences in how to define new block. Maybe a change that is less highlighted is the new feature of container blocks which implements stuff that previously was hacky to implement (like in nimislides). Maybe a bullet point or paragraph dedicated to the first feature in the overview part could be useful.

I also noticed a the them typo (I will add more details to where).

Absolutely agree on removal of `turl` or similar variable names

pietroppeter:
Found the typo ref: "This makes working with the harder"

HugoGranstrom:
Thank you, I'll add a piece about container blocks :D 

> Maybe a bullet point or paragraph dedicated to the first feature in the overview part could be useful.

Are you still referring to container blocks here or something else?

pietroppeter:
Still container blocks

HugoGranstrom:
Nice, then it should be written. Anything missing?

pietroppeter:
all good!

pietroppeter (approved):

PR #255, opened 2026-06-04
fix: gensym in newNbBlock

HugoGranstrom:
`newNbBlock` didn't handle gensym'd variables, so when you used string interpolation in `toHtml`, the variable was gensym'd but not the string interpolated variable. I fixed it by stripping the gensym part. This only occured if `newNbBlock` was called inside a template. 

```nim
toHtml:
  let x = 1
  &"{x}"
# was transformed into:
toHtml:
  let x`gensym123 = 1
  &"{x}" # doesn't match anymore
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,15 @@ Notes for maintainers:
 - When we tag a new release, we should auto generate the release notes. It does not hurt if we add more context to the release notes (e.g. taking notable elements from PR discussion). We might also want to add a release discussion post.
 - finally, after a release, we update this changelog (and bump version) using the same wording from release notes: https://github.com/pietroppeter/nimib/releases
 
+## v0.4.1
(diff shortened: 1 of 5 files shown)


HugoGranstrom:
@pietroppeter I would appreciate a quick look

pietroppeter:
I am not really competent enough on gensymmed stuff to be able to give useful advice. I guess it does not surprise me a big like this (sometimes I have to workaround some things that do not work).

I guess the change in trenders is unrelated

HugoGranstrom:
No worries 😄 It doesn't break existing code and at least fixes some non-working code. So it is a strict improvement. 

Yes it's unrelated, don't know why it didn't trigger in the last PR though 🤷

pietroppeter:
> don't know why it didn't trigger in the last PR though 🤷

image

pietroppeter (approved):

